PR #2, opened 2014-07-04
Parse if then else

lucabrunox:


jwiegley:
Nice, @lethalman!  Although we don't want to add new logic to `nixApp`.  The behavior of "reserved" keywords is that `nixApp` won't parse them as tokens.  See how I did this in `let`.

I'll make the fixes here after merging.

PR #11, opened 2014-08-13
Tests and fixes for the parser

bennofs:
This PR contains a few HUnit tests using tasty for the parser. I also fixed most of the issues I encountered while writing the tests, the only test that is still failing is a test for string parsing (which doesn't seem to be implemented fully yet). 

I also fixed a few compile errors that happened after I merged `origin/master`.  Even though there are many changes, the changes are mostly simple one line fixes.


bergey:
Looks good to me.  Thanks for catching my errors in the lamba args.  I've been meaning to ask if we have a testsuite.

jwiegley:
This is nice work, and yes, we needed to start a testsuite.

PR #12, opened 2014-08-15
parsec: Also use parsers (like trifecta)

bennofs:
This saves a lot of duplication and also makes it easier
to keep the parsec and trifecta versions of the parser in sync.


jwiegley:
Thanks!  I was wanting something like that.

PR #13, opened 2014-08-15
antiquotes + improve pretty printer + restructure

bennofs:
This pull request improves the pretty printing and adds support for
antiquotes. It also fixes an issue with the parser that caused `[if true
then false else true]` to parse successfully, even though that is not a
valid nix expression.

The pretty printer now produces a lot more readable output and also
supports operator precedences. Here is an example:

Input:

`````` nix
{ a = 3; b = 4; c = { inherit ({a = 3; }) a; e = 4; f = 5; }; d = if true then { a = 1; b = 2; c = 3; } else if false then null else { inherit ({a = 3; b = 4; cdefgads = 5; }) cdefgads; }; f = x: x; list = [ 1 2 (f 3) ((x: x) 3) [ 4 5 ] ]; }```
``````

Old:

``` nix
 { a = 3;
b = 4;
c =  { inherit ( { a = 3; }) a;
e = 4;
f = 5; };
d = if true
then  { a = 1;
b = 2;
c = 3; }
else if false
then null
else  { inherit ( { a = 3;
b = 4;
cdefgads = 5; }) cdefgads; };
f = x:
x;
list = [ 1 2 f (3) x:
x (3) [ 4 5 ] ]; }
```

New:

``` nix
{
  a = 3;
  b = 4;
  c = {
    inherit ({ a = 3; }) a;
    e = 4;
    f = 5;
  };
  d = if true
    then { a = 1; b = 2; c = 3; }
    else if false
      then null
      else {
        inherit ({
          a = 3;
          b = 4;
          cdefgads = 5;
        }) cdefgads;
      };
  f = x: x;
  list = [
    1
    2
    (f 3)
    ((x: x) 3)
    [ 4 5 ]
  ];
}
```

---
### The changes to the AST are:
- strings are no longer atomic, because they may contain other
  expressions in the form of antiquotes. For strings, the new type
  NString is introduced and the constructor NStr is added to NExprF
- the unused NVar constructor of NExprF is removed
- operators are now represented uniformly so that the pretty printer
  can lookup information about operators (in particular, associativity
  and precedence)
- the NArgs constructor is removed. The first argument of the NAbs
  constructor now directly represents the lambda arguments.
- the select and the hasattr operator are moved into NExpr because
  they are special (they only accept a selector as second argument, and
  select also supports 'or')
### Other changes

The list of operators is now in Types.hs and Parser.hs re-uses that list
to build the parser. This is required because the pretty printer and
parser both need access to operator precedences.

Parser and evaluator also support dynamic attributes and attributes with
dots now. As an example, `let b.a = 3; b.c = { e = {}; }; b.c.e.${"f"} =
4; in b` is parsed and evaluated correctly. As a side effect, NSym
values now don't evaluate to themselves anymore, but instead to the
value retrieved by looking up the variable in the current environment.

Support for evaluating `inherit` bindings was removed because it was
broken before (`{ inherit a; }` would evaluate to a set where the
attribute `a` had the value `NSym a`, not the value of `a`).

The manual Show instances for the AST were replaced by derived
ones, because the manual instances often resulted in output were it was
difficult to determine the missing parentheses.

I'm sorry that this PR got so big, I started working on a better pretty printer, but while I was changing the AST for that, I mixed that work with support for antiquotations. Combined with some fixes in the places I changed, that turned out to be a very big change. If it's important, I could try to split the commit.

### DIFF
--- a/Nix.hs
+++ b/Nix.hs
@@ -17,6 +17,7 @@ nix path = do
         Success n -> do
             displayIO stdout $ renderPretty 0.4 80 (prettyNix n)
             top <- evalExpr n (Fix (NVSet Map.empty)) -- evaluate top level
+            putStrLn ""
             print top
 
 main :: IO ()
--- a/Nix/Eval.hs
+++ b/Nix/Eval.hs
@@ -2,55 +2,54 @@ module Nix.Eval where
 
 import           Control.Applicative
 import           Control.Arrow
-import           Control.Monad hiding (mapM)
+import           Control.Monad hiding (mapM, sequence)
+import           Data.Foldable (foldl')
 import qualified Data.Map as Map
 import           Data.Text (Text)
+import qualified Data.Text as Text
 import           Data.Traversable as T
 import           Nix.Types
-import           Prelude hiding (mapM)
+import           Prelude hiding (mapM, sequence)
 
-buildArgument :: NValue -> NValue -> NValue
-buildArgument paramSpec arg =
-    -- Having the typed lambda calculus would make this code much safer.
-    Fix $ NVSet $ case paramSpec of
-        Fix (NVArgSet s) ->
-            case arg of
-                Fix (NVSet s') ->
-                    Map.foldlWithKey' (go s') Map.empty s
-                _ -> error "Unexpected function environment"
-        Fix (NVConstant (NSym name)) -> Map.singleton name arg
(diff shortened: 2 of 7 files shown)


jwiegley:
Thanks!  Btw, previously I was representing anti-quoted strings simply as "lists that can coerce to a string".

PR #17, opened 2014-09-11
null is a constant added to the base environment, not a reserved word

shlevy:
Opening as a PR to get @jwiegley's opinion.


jwiegley:
@shlevy I'm totally fine with this, happy even.

PR #19, opened 2015-06-23
adding OverloadedStrings extension, hiding conflicting <$> operator

adnelson:
@jwiegley This fixes some errors I was getting when attempting to use GHCI with these modules. Parenthetically, it's pretty lame that the `Text.PrettyPrint` module decided to make their own incompatible version of `<$>`... I guess maybe that was before it became a common `fmap` synonym though.

Also a minor grammatical fix.


jwiegley:
@adnelson Thanks!

PR #20, opened 2015-06-23
added a bunch of wrapper functions, simpler representation of empties

adnelson:
@jwiegley I extended the current crop of wrapper functions to make `NExpr`s. Also added some functions for manipulating `NExpr`s: specifically, adding additional bindings to `NLet`s and `NSet`s, and applying a function to the body of an `NAbs`. Conceivably, there could be a whole library of these kinds of transformations to facilitate programmatic manipulation of nix expressions. It's probably a bit better to do that stuff through lenses, but as I don't know much about lens, I just left them as straightforward partial functions.

### DIFF
--- a/Nix/Pretty.hs
+++ b/Nix/Pretty.hs
@@ -119,16 +119,14 @@ prettyNix = withoutParens . cata phi where
   phi :: NExprF NixDoc -> NixDoc
   phi (NConstant atom) = prettyAtom atom
   phi (NStr str) = simpleExpr $ prettyString str
+  phi (NList []) = simpleExpr $ lbracket <> rbracket
   phi (NList xs) = simpleExpr $ group $
     nest 2 (vsep $ lbracket : map (wrapParens appOpNonAssoc) xs) <$> rbracket
+  phi (NSet rec []) = simpleExpr $ recPrefix rec <> lbrace <> rbrace
   phi (NSet rec xs) = simpleExpr $ group $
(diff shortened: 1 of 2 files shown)


jwiegley:
If you can describe the manipulations you'd like to perform, I'd be happy to write the first few lenses for you.  It's a rather straightforward pattern.  Mainly they are all functions of this shape:

``` haskell
someLens f x =
    -- x has the incoming "structure" s, while f maps a -> f b; we need to return f t, so:
    wrapAround x <$> f (extractFrom x)

    -- extractFrom pulls the 'a' from 's'; wrapAround builds a 't' around 'b'
```

Traversals will look almost exactly like an implementation of `traverse` for the `Traversable` class, except that you get to pick the type parameter you're traversing over.

Prisms add `Maybe` to the lens pattern, so if `x` doesn't have the right shape, you can indicate that.

jwiegley:
Also, I'm OK with fixing GHC at 7.8, so we could use pattern synonyms to provide specialized constructors, rather than a series of `mk*` functions.  This has the added benefit of making pattern matching code on these constructions cleaner whenever that's necessary.

adnelson:
Well you can pretty much see the manipulations I'm doing (there are two that I'm using for `nixfromnpm`). The first applies a function to the body of a nix function, and the second appends additional bindings to either a set or a let.

Pattern synonyms sound great.

Issue #16, opened 2014-08-25
Switch parser library to megaparsec


bennofs:
The parser is currently quite slow compared to the C parser:

```
$ time nix-instantiate --parse ~/nixpkgs/pkgs/top-level/all-packages.nix > /dev/null
0.10user 0.02system 0:00.36elapsed 33%CPU (0avgtext+0avgdata 8908maxresident)k
0inputs+0outputs (0major+2511minor)pagefaults 0swaps
```

compare that to hnix:

```
$ time dist/build/hnix/hnix ~/nixpkgs/pkgs/top-level/all-packages.nix  > /dev/null 
hnix: Operators are not yet defined

real    0m6.914s
user    0m4.056s
sys 0m0.172s
```

I don't know how to improve performance though.

Mathnerd314:
Happy is much closer to the Bison parser that Nix uses, and probably of comparable speed. Does anyone mind if I switch hnix away from trifecta/parsec?

jwiegley:
I do.  We just need to optimize the present parser.  I think many more people are comfortable with reading and editing Parsec, than they are using Happy.  So, for a speed gain, we'd reduce our pool of candidate contributors. I'm pretty sure we can get there with Trifecta, we probably just have some nasty backtracking happening in a few key places.

Mathnerd314:
Hmm. Well, I for one am not sure how to get Parsec/Trifecta to parse the `let-body` syntax:
`let { a = 1; body = 2; }`
This is a perfectly fine Nix expression, but HNix expects it to be of the form `let { ... } in ...`.

Happy has a GLR mode which seems to be working, although it is proving difficult to integrate it with the rest of hnix.

cstrahan:
> Hmm. Well, I for one am not sure how to get Parsec/Trifecta to parse the let-body syntax:
> let { a = 1; body = 2; }

Wow, I didn't know that was legal syntax. Crazy.

steshaw:
The performance with respect to the C parser seems to have improved:

```
$ time nix-instantiate --parse ~/.nix-defexpr/channels/nixpkgs/pkgs/top-level/all-packages.nix  >/dev/null

real	0m0.139s
user	0m0.120s
sys	0m0.010s
```

```
$ time dist/build/hnix/hnix ~/.nix-defexpr/channels/nixpkgs/pkgs/top-level/all-packages.nix >/dev/null

real	0m1.306s
user	0m1.236s
sys	0m0.063s
```

[...]

PR #22, opened 2015-06-23
not evaluating in hnix executable, only parsing

adnelson:
@jwiegley The eval function is not complete and produces errors (e.g. when trying out #16). This removes that and also adds in a basic usage string.


jwiegley:
Yes, eval is really where work needs to happen next.

I'm giving you push privileges.

PR #27, opened 2015-10-15
Minor fixes in pretty printing.

adnelson:
- Correctly escape double quotes. 
- Use quotes when bindings are reserved words.

### DIFF
--- a/Nix/Parser/Library.hs
+++ b/Nix/Parser/Library.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Nix.Parser.Library ( module Nix.Parser.Library, module X) where
 
+import Prelude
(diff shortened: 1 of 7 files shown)


Nix/Parser/Library.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Nix.Parser.Library ( module Nix.Parser.Library, module X) where
 
+import Prelude
```
bennofs, line 6: Hmm, why is this neccessary? If it's to avoid AMP warnings, I thought that only works well if the import of Prelude is the last import?

jwiegley, line 6: I'm not sure it is necessary at all.

PR #28, opened 2015-10-28
Fix quote escaping

adnelson:
Hey @jwiegley I can merge this in but can you upload the latest to hackage?

### DIFF
--- a/Nix/Eval.hs
+++ b/Nix/Eval.hs
@@ -98,7 +98,7 @@ evalExpr = cata phi
 evalString :: NValue -> NString (NValue -> IO NValue) -> IO Text
 evalString env (NString _ parts)
   = Text.concat <$> mapM (runAntiquoted return (fmap valueText . ($ env))) parts
-evalString env (NUri t) = return t
+evalString _ (NUri t) = return t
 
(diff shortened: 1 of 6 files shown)


Nix/Eval.hs
```diff
@@ -98,7 +98,7 @@ evalExpr = cata phi
 evalString :: NValue -> NString (NValue -> IO NValue) -> IO Text
 evalString env (NString _ parts)
   = Text.concat <$> mapM (runAntiquoted return (fmap valueText . ($ env))) parts
-evalString env (NUri t) = return t
+evalString _ (NUri t) = return t
```
adnelson, line 101: This and the other change in this file are just to suppress compiler warnings.

Nix/Types.hs
```diff
@@ -48,9 +48,17 @@ atomText :: NAtom -> Text
 atomText (NInt i)  = pack (show i)
 atomText (NBool b) = if b then "true" else "false"
 atomText NNull     = "null"
-atomText (NPath s p)
-  | s = pack ("<" ++ p ++ ">")
-  | otherwise = pack p
+atomText (NPath isFromEnv p)
```
adnelson, line 51: This is so a path like `mkPath False "foo"` doesn't get rendered as just `foo`, but instead as `./foo`

Issue #33, opened 2016-03-11
Make parse round-trip-able


cstrahan:
There are a number of tools (including one I desperately need to write) that would require the ability to parse a nix expression, transform it in some way, and then write the expression back out - all while keeping unmodified portions unchanged in the output.

I could take a stab at this, but I think I might need some hints/help.

jwiegley:
That's a great idea @cstrahan.  You'll need to plumb attributes through the syntax tree that preserve knowledge like surrounding whitespace, etc.  This part should be fairly easy, since we're using a `Free` representation.

Then, in the evaluator, you only need to preserve the attributes over any transformation (they would be abstract in that case). In the pretty printer, it would use the attributes for laying out the elements to be printed, or any other style decisions.

It's not exactly trivial, but should be straightforward, and I'm sure there are examples of something similar living out on the Web.

cstrahan:
@jwiegley I don't think trailing white space matters as much, so perhaps just tracking source spans would be sufficient (ala `haskell-src-exts`)?

If so, do you suppose I should add a monomorphic field (e.g. `SrcSpan`) to each construct? `haskell-src-exts` uses a polymorphic field in their annotated AST, but I don't know if we'd benefit from going that route.

jwiegley:
You shouldn't need to change the AST at all, but just change the `Free` type synonym.

cstrahan:
Do you mean the `Fix` synonym? (grepping for `Free` doesn't turn up anything)

Are you suggesting something like this?

``` Haskell
data SrcSpan = SrcSpan Int Int
data Ann x f a = Ann x (f a)
type NExpr = Fix (Ann SrcSpan NExprF)
```

jwiegley:
Oh yeah, sorry, `Fix`.  And yes, what you pasted is exactly what I had meant.  That will give you an annotation at every level of the resulting tree.

expipiplus1:
I've started working towards this in #35

jwiegley:
@cstrahan We've now merged @expipiplus1 's work, does this help you in building your tools?

cstrahan:
@jwiegley I believe so, although I've shelved that work for the time being. Thanks!

expipiplus1:
For an example of this being used you can take a look here: https://github.com/expipiplus1/update-nix-fetchgit

This doens't do a round trip though, It uses the source locations to modify the underlying string.

PR #35, opened 2016-06-23
WIP: Annotate expressions with their location

expipiplus1:
Some initial work towards fixing #33 

This is far from complete, Some functions need implementing, and there's some tidying up to do in the parser. `Annotation.hs` could also use a little more organisation.

It typechecks though :)

### DIFF
--- a/Nix/Expr.hs
+++ b/Nix/Expr.hs
@@ -1,8 +1,10 @@
 -- | Wraps the expression submodules.
 module Nix.Expr (
   module Nix.Expr.Types,
+  module Nix.Expr.Types.Annotated,
   module Nix.Expr.Shorthands
   ) where
 
 import Nix.Expr.Types
 import Nix.Expr.Shorthands
+import Nix.Expr.Types.Annotated
--- a/Nix/Expr/Shorthands.hs
+++ b/Nix/Expr/Shorthands.hs
@@ -1,4 +1,7 @@
 -- | A bunch of shorthands for making nix expressions.
+--
+-- Functions with an @F@ suffix return a more general type without the outer
+-- 'Fix' wrapper.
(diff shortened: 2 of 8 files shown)


jwiegley:
@expipiplus1 Nice! If this is a work in progress, shouldn't I wait until you're done?

expipiplus1:
Yes! Please don't merge now!

expipiplus1:
@jwiegley How would you suggest dealing with whitespace in the expression annotations?

It's not possible to eat the whitespace before recording the start of the span as some parsers are whitespace sensitive (the tests fail for `select`).

There isn't any opportunity to record trailing whitespace, as `token` eats any trailing whitespace.

The best option I can think of is to postprocess the source spans to remove leading and trailing whitespace.

Another option would be to split the parser and the lexer, but that'd probably be a much larger job.

Another option would be to leave it as is, and let the user deal with it :)

jwiegley:
For now we can let the user deal with it, but I like the idea of splitting parser and lexer.

expipiplus1:
The annoying thing is that _I'm_ the user at the moment :)

If I find a tidy way of stripping whitespace I think I'll do that.

expipiplus1:
Another possibility would be to modify the definition of whitespace so that it maintains some kind of `beforeWhitespacePos` + `afterWhitespacePos` state. I'm not sure if this is possible in a Trifecta parser though. Does this idea makes sense?

expipiplus1:
This is working well enough now for me to have made this little utility to automatically update `fetchgit` values in nix expressions to the latest versions.

https://github.com/expipiplus1/update-nix-fetchgit

expipiplus1:
@jwiegley aside from some tests and documentation, is there anything major preventing this from being merged

PR #42, opened 2016-09-21
Add Show1 instances for base 4.9

expipiplus1:
Due to differences in how `Show1` is defined this is not compatible with `base < 4.9`. Someone should probably do some CPP work for that to work if supporting older versions of GHC is desired.


expipiplus1:
hnix compiles without a problem on `GHC-8.0.1` with this change.

jwiegley:
This is great, although I'm not ready to force 8.0.1 on everyone just yet.  So give it a few months, and then we'll merge this and not bother with CPP markers, unless someone really needs them.

expipiplus1:
Sounds good.

jwiegley:
Now required as of 0.3.3.  On to the future!

Issue #58, opened 2017-05-20
string pretty printer escapes `$` when it's not desired


Fuuzetsu:
```
                , Nix.bindTo "installPhase" $ Nix.mkIndentedStr $ Text.unlines $
                    [ "mkdir $out"
                    , "cp $src $out"
                    ]
```
produces

```
        installPhase = ''
          mkdir ''$out
          cp ''$src ''$out
        '';
```

There seems to be no work-around as similar issue happens with `mkStr` too. What's the intended way to generate strings that point at variables at run-time?

bennofs:
```
nix-repl> '' $foo ''
"$foo "

nix-repl> '' ''$foo ''
"$foo "
```

They seem to evaluate to the same thing?

Fuuzetsu:
Hm, interesting. I wrote a work-around in meantime for my use-case but maybe it wasn't needed.

```
          mkUnescapedIndentedStr rep s =
            Nix.mkSym "builtins.replaceStrings"
              `Nix.mkApp` Nix.mkList [Nix.mkStr rep]
              `Nix.mkApp` Nix.mkList [Nix.mkSym "\"$\""]
              `Nix.mkApp` Nix.mkIndentedStr s
```

```
                , Nix.bindTo "installPhase" $ mkUnescapedIndentedStr "ENV_"
                    $ Text.unlines [ "mkdir ENV_out"
                                   , "cp ENV_src ENV_out"
                                   ]
```
```
        installPhase = builtins.replaceStrings [
          "ENV_"
        ] [ "$" ] ''
          mkdir ENV_out
          cp ENV_src ENV_out
        '';
```

@bennofs there is still a bug here however

```
nix-repl> out = "/tmp"

nix-repl> '' echo ${toString out} ''
"echo /tmp "

nix-repl> '' echo ''${toString out} ''
"echo ${toString out} "
```

bennofs:
@Fuuzetsu I don't think this is a bug. What you pass to `mkStr` is a plain string, so every nix special char should be escaped. The correct expression for an interpolated string would be something like this:

```
λ: parseNixString " '' ${out} '' "
Success (Fix (NStr (Indented [Antiquoted (Fix (NSym "out")),Plain " "])))
λ: prettyNix $ Fix (NStr (Indented [Antiquoted (mkSym "out"), Plain " foo"]))
''${out} foo''
```

Fuuzetsu:
[...]

PR #59, opened 2017-05-31
Dollar quoting

rimmington:
The Nix manual [states]:

> Since ${ and '' have special meaning in indented strings, you need a way to quote them. ${ can be escaped by prefixing it with '' (that is, two single quotes), i.e., ''${.

but says nothing about needing to escape `$` followed by any other character. hnix currently escapes `$` unconditionally when pretty-printing, which is also accepted by the Nix tooling:

```
λ> show . prettyNix $ Fix (NStr (Indented [Plain "$foo"]))
"''''$foo''"
```
```
nix-shell> ''''$foo''
"$foo"
```

I can't find any documentation about this behaviour by the Nix tools.

None of this is really a _bug_, but it does make the common case of including shell variables in Nix strings a bit unintuitive; at first glance I thought this string ended too early:

```
preConfigure = ''
  export PATH="''$PWD/dist/build/intero:''$PATH"
'';
```

So this PR modifies the pretty-printer to only escape `${` in indented strings. I would've changed the printing of non-indented strings as well, but I can't find where that escaping is done.

Somewhat related to #58.

[states]: https://nixos.org/nix/manual/#idm140737318139904

### DIFF
--- a/Nix/Pretty.hs
+++ b/Nix/Pretty.hs
@@ -70,7 +70,7 @@ prettyString (Indented parts)
     f ([Plain t] : xs) | Text.null (strip t) = xs
     f xs = xs
   prettyLine = hcat . map prettyPart
-  prettyPart (Plain t) = text . unpack . replace "$" "''$" . replace "''" "'''" $ t
+  prettyPart (Plain t) = text . unpack . replace "${" "''${" . replace "''" "'''" $ t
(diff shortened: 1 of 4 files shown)


Fuuzetsu:
This actually solves half of #58, the other half is analogous fix for non-indented strings.

PR #64, opened 2017-09-23
Reduce code duplication in main/Main.hs

vaibhavsagar:
`nix` and `nixString` do mostly the same thing.


vaibhavsagar:
Thank you :smile:

jwiegley:
@vaibhavsagar Thank you! Both for noticing the duplication, and providing a fix. :)

PR #74, opened 2018-01-18
shell.nix: pin nixpkgs revision to HEAD of nixos-17.09

eqyiel:
This is just for the sake of people not on the stable channel who want to use
the same version of GHC specified here.

### DIFF
--- a/nixpkgs.nix
+++ b/nixpkgs.nix
@@ -1,0 +1,26 @@
+{ stdenv, fetchFromGitHub }:
+
+stdenv.mkDerivation rec {
+  name = "nixos-17.09-${version}";
+  version = "2017-01-19";
+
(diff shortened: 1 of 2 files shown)


eqyiel:
This actually fails for me: https://gist.github.com/eqyiel/8df65e30ae2f54ddd40172ba150e8ecb

I don't know why it's pulling in `ghc-7.4.2` in addition to `ghc-8.0.1`.

@jwiegley I'm curious to know what revision of nixpkgs you are on in order to make this work!

jwiegley:
I'm at 3a4698f8136 right now.

jwiegley:
Note that I'm going to remove `default.nix` and `shell.nix` from this repository, since they can both be generated by `cabal2nix`.

PR #76, opened 2018-01-26
Fix recset evaluation logic

madjar:
Before, an evaluation of a recset would contain all the environment. Eg:
```
>>> let a = 4; in rec {x = 2;}
{x = 2; a = 4;}
```

This commit changes it to only contain the values actually bound in the set:
```
>>> let a = 4; in rec {x = 2;}
{x = 2;}
```

### DIFF
--- a/Nix/Eval.hs
+++ b/Nix/Eval.hs
@@ -148,7 +148,7 @@ evalExpr = cata phi
         rec
           mergedEnv <- pure $ Fix $ NVSet $ evaledBinds `Map.union` env'
           evaledBinds <- evalBinds True mergedEnv binds
-        pure mergedEnv
+        pure . Fix . NVSet $ evaledBinds
       _ -> error "invalid evaluation environment"
 
     phi (NLet binds e) = \env -> case env of


jwiegley:
Great! Speaking of evaluation environments, any thoughts on this one:

https://github.com/jwiegley/hnix/blob/master/Nix/Eval.hs#L194

madjar:
Took me a while to figure out what was meant in that comment, so I'll write it down for the records:

```
let f = {a, b ? a}: {inherit a; inherit b;}; in f {a = 2;}
```

In nix, the default values for arguments of a lambda can refer to other arguments (I did not know that :)).

So, now that it's clearer to me, I'll think about how this can be implemented here.

madjar:
Okay, so my conclusion is that `NVFunction (Params r) (NValue m -> m r)` is not a good representation, because it does not leave room for the fact that you need the arguments from the function application to compute the default values.

I'm working on a solution that changes it to `NVFunction (Params (NValue m -> m r)) (NValue m -> m r)`, so that the default values depend from the application arguments, and not only from the environment.

I'll send a PR for that, but there's a thing or two I'd like to tackle before.

jwiegley:
This sounds great, @madjar.  I know that @ryantrinkle has been blocked on this issue for a while.

ryantrinkle:
Yes, thanks @madjar and @jwiegley for looking at this!

PR #78, opened 2018-01-28
Change the environment in evaluation to be a `Map Text (NValue m)`

madjar:
Currently, the environment is passed as a `NValue m`, but is assumed to be
a set every single time it is used. This commit changes it `Map Text (NValue
m)`. Since this is used a lot, it defines a new type alias:

```
type Set m = Map.Map Text (NValue m)
```

This has multiples benefits:
- Simplify some code by removing all the checks that the env is indeed a set
- Simplify the usage of the module by making clear that we need a set as the
  environment. (I especially like this, since it took me a while to figure out
  what that argument was supposed to be the first time).
- Make it simple to inject functions in the environment (for example to have
  builtins) since now the function definition doesn't need to unwrap the set.

### DIFF
--- a/Nix/Eval.hs
+++ b/Nix/Eval.hs
@@ -27,7 +27,7 @@ data NValueF m r
     | NVStr Text
     | NVList [r]
     | NVSet (Map.Map Text r)
-    | NVFunction (Params r) (NValue m -> m r)
+    | NVFunction (Params r) (ValueSet m -> m r)
     | NVLiteralPath FilePath
     | NVEnvPath FilePath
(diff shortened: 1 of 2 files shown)


jwiegley:
`Set` is a bit of an overload, if we ever choose to use `Data.Set` in this library. Maybe another name, like `ValueSet`?

madjar:
Yeah, I was not super satisfied with the name, `ValueSet` sound better.

madjar:
There, I changed it to `ValueSet` (and fixed the conflict).

PR #79, opened 2018-01-28
Add a pretty-printer for NValue

madjar:
Here is a little thing I wrote writing a small repl with hnix. As explained in the commit message, I used a trick for the implementation, so feel free to reject this PR if you don't like the way this is implemented.

-----

The implementation is a bit hacky: we convert the NValue to a NExpr, which
let us reuse the ast pretty-printing logic.

This commit also moves `atomText` from `Nix.Pretty` to `Nix.Eval` to avoid a
circular dependency. I don't have a strong argument for that, except that the
pretty-printer depending from the evaluator makes a bit more sense than other
way around.

### DIFF
--- a/Nix/Eval.hs
+++ b/Nix/Eval.hs
@@ -15,7 +15,6 @@ import qualified Data.Text as Text
 import           Data.Traversable as T
 import           Data.Typeable (Typeable)
 import           GHC.Generics
-import           Nix.Pretty (atomText)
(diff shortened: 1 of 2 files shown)


jwiegley:
I like it.  Maybe this points to the need for a third, common module between the two?

madjar:
Oh, I just realized there is a `Nix.Atom` module that would be a perfect fit for this.

PR #82, opened 2018-02-09
Run nix language tests from upstream

layus:
At the moment, most tests fail, mainly due to missing builtins
It is nonetheless a good framework to start improving hnix.

### DIFF
--- a/Nix/Builtins.hs
+++ b/Nix/Builtins.hs
@@ -1,0 +1,97 @@
+
+module Nix.Builtins (baseEnv, builtins, evalTopLevelExpr) where
+
+import           Control.Applicative
+import           Control.Monad hiding (mapM, sequence)
+import           Control.Monad.Fix
+import           Data.Fix
+import qualified Data.Map as Map
+import           Data.Text (Text)
+import qualified Data.Text as Text
+import           Data.Traversable (mapM)
+import           Nix.Atoms
+import           Nix.Eval
+import           Nix.Expr (NExpr)
+import           Prelude hiding (mapM, sequence)
+
+-- | Evaluate a nix expression in the default context
+evalTopLevelExpr :: MonadFix m => NExpr -> m (NValue m)
+evalTopLevelExpr val = evalExpr val baseEnv
+
+baseEnv :: MonadFix m => ValueSet m
+baseEnv = Map.fromList $ [ ("builtins", Fix $ NVSet builtins) ] ++ topLevelBuiltins
+  where
+    topLevelBuiltins = map mapping $ filter isTopLevel builtinsList
+    -- builtins = Map.fromList $ map mapping $ builtinsList
+
+builtins :: MonadFix m => ValueSet m
+builtins = Map.fromList $ map mapping $ builtinsList
+
+data BuiltinType = Normal | TopLevel
+data Builtin m = Builtin {kind :: BuiltinType, mapping :: (Text, NValue m) }
+
+isTopLevel :: Builtin m -> Bool
+isTopLevel b = case kind b of
+                    Normal -> False
+                    TopLevel -> True
+
+builtinsList :: MonadFix m => [ Builtin m ]
+builtinsList = [
+      topLevel ("toString", prim_toString)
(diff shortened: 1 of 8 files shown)


jwiegley:
This is fantastic, thank you!

jwiegley:
There are a lot of other changes confounded with this pull request. I like the idea of integrating the Nix language tests, but let's do that as a separate thing. It's hard to review what you've done with these other changes in there as well.

layus:
There is indeed a lot of related changes that may be split into different pull requests.

But it seemed to me that the individual commits where easy enough to review individually. Making the PR reviewable as a set of small commit-wise PRs.

I may split it later when I find the time to do so.

layus:
Also, I was hoping for some feedback on the implementation before cleaning things up. Especially on NixLangTests.hs and Builtins.hs. I am quite new to Haskell.

layus:
Voilà, this PR has been split and reduced to its minimal form.

madjar:
Hi there, this looks pretty awesome!

I'm wondering (but that probably for some future refactoring) if it's possible to rework `NVFunction` so that `NVBuiltin1` doesn't have to be a separate case, and so that `NVBuiltin2` can be implemented with currying of `NVFunction` rather than as a separate case.

Nix/Builtins.hs
```diff
@@ -0,0 +1,117 @@
...
+
+-- | Evaluate a nix expression in the default context
+evalTopLevelExpr :: MonadFix m => NExpr -> m (NValue m)
+evalTopLevelExpr val = evalExpr val baseEnv
+
+baseEnv :: MonadFix m => ValueSet m
+baseEnv = Map.fromList $ [ ("builtins", Fix $ NVSet builtins) ] ++ topLevelBuiltins 
+  where
+    topLevelBuiltins = map mapping $ filter isTopLevel builtinsList
+    -- builtins = Map.fromList $ map mapping $ builtinsList
```
madjar: Some leftovers comments here?

Nix/Builtins.hs
```diff
@@ -0,0 +1,117 @@
...
+
+data BuiltinType = Normal | TopLevel
+data Builtin m = Builtin {kind :: BuiltinType, mapping :: (Text, NValue m) }
+
+isTopLevel :: Builtin m -> Bool
+isTopLevel b = case kind b of 
+                    Normal -> False
+                    TopLevel -> True
+
+--------------
```
madjar: It looks like some of this is implemented elsewhere. Can you remove what's been moved, and maybe add a comment explaining why the rest is there?

layus: It is implemented right below :-)

Nix/Builtins.hs
layus: yes, again implemented right below :-)

layus:
@madjar There are ways to avoid using different builtins, for example see https://stackoverflow.com/questions/16589145/typeclass-for-functions-with-different-numbers-of-arguments and http://okmij.org/ftp/Haskell/polyvariadic.html#polyvar-fn. This is a bit too far-fetched for me ATM.

There must also be a cleaner way to define the builtins without defining two function each time, but the right way to do this eludes me.

Anyway, thanks for reviewing !

madjar:
Thinking out loud: considering we have
```hasAttr :: MonadFix m => NValue m -> NValue m -> m (NValue m)```

Could we define `prim_hasAttr` as:

```
builtin2 name function = Fix $ NVBuiltin1 name (\arg -> Fix $ NVBuiltin1 name (function arg))
prim_hasAttr = builtin2 "hasAttr" hasAttr
```

That way, calling `hasAttr` with one argument returns an `NVBuiltin1` that has the argument in its closure, rather than having it explicitly in a list. As a return, we only need `NVBuiltin` with no numbers :)

layus:
@madjar Yes, I had the same idea but somehow I never managed to get the types right. It failed at invocation time in Eval.hs, for some obscure typing reasons that I could not debug or failed to understand.

I would be glad to learn from external improvements on this code

PR #87, opened 2018-02-16
Add context to strings

madjar:
This commit adds a `DList Text` to the NVStr constructor. This allows haskell
code that generates string to give them a context, which will be tracked when
the string in used.

This allows, for example, to determine the dependencies of a derivation.

-------

I used a DList, because I want O(1) concatenation, and I don't expect to read it except one when I consume it to find out the dependencies of a computation.

## Alternative

Instead of hardcoding for `DList Text`, we could make it work with any monoid. That would adding one more type parameter to all types (`NValueF c m r`, `NValue c m` and `ValueSet c m`)

### DIFF
--- a/Nix/Builtins.hs
+++ b/Nix/Builtins.hs
@@ -67,18 +67,18 @@ evalPred pred = error $ "Trying to call a " ++ show pred
 prim_toString :: MonadFix m => Functor m => NValue m
 prim_toString = Fix $ NVBuiltin1 "toString" $ toString
 toString :: MonadFix m => NValue m -> m (NValue m)
-toString s = return $ Fix $ NVStr $ valueText s
+toString s = return $ Fix $ uncurry NVStr $ valueText s
 
 prim_hasAttr :: MonadFix m => NValue m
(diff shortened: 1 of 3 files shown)


jwiegley:
@madjar Looks good. How about instead of using a new dlist dependency, we just use `Endo`?  As long as you don't need to apply other sorts of list operations on the dlists, then Endo works nicely for simply accumulating strings in this style.

madjar:
@jwiegley I'm not entirely convinced. The signature `NVStr Text (Endo ([Text]))` seems much less clear to me, so I'm afraid that while saving us a dependency, it would have a non-trivial cost in term of readability and understandability of the code.

The call for this trade-off is your, though, and you get the final say :)

jwiegley:
@madjar Well that's easy. Elsewhere just define:

``` haskell
type DList a = Endo [a]
```

madjar:
@jwiegley Here you go :)

One change I made while rebasing is changing the return type of `NVLiteralPath + NVStr` from `NVLiteralPath` to `NVStr`, in order not to discard the context. For reference, the behaviour of C++nix for this is to fail if the context is non-empty ; I can do that, but then I'll move away from DList.

PR #88, opened 2018-02-24
Simplify mkDots, handle some edge cases

adnelson:
* Remove unnecessary partiality in `printNix` function
* Remove some unused imports
* For the `mkDots` helper function, a few tweaks:
  * Don't apply an empty list of attributes
  * If the expression being dotted is itself an `NSelect`, just extend the list of selectors
  * There doesn't seem to be any need to switch between static and dynamic keys based on the contents of the key, since the pretty-printer will apply quotes if necessary
* Pretty-printer should treat an `NSelect` which has no attributes as being the same as the internal expression
* Handle an edge case of an empty-string selector in an `NSelect`

### DIFF
--- a/Nix/Expr/Shorthands.hs
+++ b/Nix/Expr/Shorthands.hs
@@ -6,10 +6,11 @@ module Nix.Expr.Shorthands where
 
 import Prelude
 import Data.Monoid
-import Data.Text (Text)
+import Data.Text (Text, unpack)
 import Data.Fix
 import qualified Data.Map as Map
 import Nix.Atoms
+import Nix.Parser.Library (reservedNames)
 import Nix.Expr.Types
 import Text.Regex.TDFA.Text ()
 import Text.Regex.TDFA ((=~))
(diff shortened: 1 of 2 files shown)


jwiegley:
Thanks!

madjar:
@adnelson I'm looking into the CI failures, and it seems `mkDotsSymbolEscaping` switched to red after this PR. For the values `[ "abc def", "\\", "'abc", "\"", "-foo", "a.b.c" ]`, the test expect the result of `mkDot` to be of the form `DynamicKey (Plain (DoubleQuoted [Plain _]))` (which won't happen). Can you confirm whether the tests are wrong? Thanks!

adnelson:
Yeah, it just uses `StaticKey` for everything. The idea is that the nix AST doesn't care whether a key name is empty or alphanumeric etc; it's only when it needs to pretty-print it into a string that the need for quoting arises. 

I attempted to run tests locally but saw a large number of failures unrelated to my change, so I assumed that the code was just in a non-functional state. Should have looked for a test of this function :) I'd be happy to update it.

madjar:
Yeah, the tests now include running all the tests from the C++ implementation, some of which fail :). I run the tests with `stack test --ta '-p !upstream'` to exclude them. I'm trying to implement a "don't fail for expected failures" for them.

@adnelson If you can fix this test, that'd be cool!

PR #89, opened 2018-02-27
Eval: coerce a set to a string if there is a __asString attribute

madjar:
This is the last missing piece for my "nix for data jobs" prototype. It allows a set to be coerced to a string if it has a `__asString` attribute.

For reference, in the C++ nix, this can happen in two ways:
- If the set has an `outPath` attribute (in which case it is used);
- If the set has a `__toString` attribute, which is a function (in which case it is called with the set as argument).

I went for `__asString`, as it makes sense outside of the specific context of implementing derivations in nix.

### DIFF
--- a/Nix/Eval.hs
+++ b/Nix/Eval.hs
@@ -67,7 +67,9 @@ valueText = cata phi where
     phi (NVConstant a)    = (atomText a, mempty)
     phi (NVStr t c)       = (t, c)
     phi (NVList _)        = error "Cannot coerce a list to a string"
-    phi (NVSet _)         = error "Cannot coerce a set to a string"
+    phi (NVSet set)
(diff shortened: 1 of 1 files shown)


adnelson:
Personally I’d prefer to follow what’s in the reference implementation. If it uses __toString, I think we should too.

madjar:
I was going to argue that we don't have to follow undocumented implementations details, but the behaviour of `__toString` actually documented, so I stand corrected :)

For my use case, what I actually need is the behavior of `outPath` (which is what is implemented in this PR), which is an implementation detail in nix. Thus, I propose naming it `__asString` (or `__stringValue`) rather than `outPath` because the later makes things confusing.

I agree that honoring `__toString` would also be nice, and I can do it in a future PR.

PR #90, opened 2018-02-27
Represent builtins with 2 arguments with currying

madjar:
As a result, `NVBuiltin2` is not needed anymore, and `NVBuiltin1` gets renamed
to `NVBuiltin`

@layus This is what we were talking about in #82, if you're interested.

### DIFF
--- a/Nix/Builtins.hs
+++ b/Nix/Builtins.hs
@@ -65,18 +65,18 @@ evalPred pred = error $ "Trying to call a " ++ show pred
 -- Primops
 
 prim_toString :: MonadFix m => Functor m => NValue m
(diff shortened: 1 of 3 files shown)


layus:
Nice! I somehow failed to achieve this.

Is there a way to make hunit ignore the failed tests (but still log them ?)
The travis CI now fails all the time due to the added tests.

madjar:
@layus I was looking at that yesterday. We could try splitting the travis targets between the upstream tests and the not-upstream tests, in order to get separated reporting, but I didn't go all the way to figuring how to make the `default.nix` run only some of the tests.

(Incidentally, `ShorthandTests` is now failing)

PR #91, opened 2018-02-28
remove mkDots test, no longer necessary

adnelson:
@madjar I felt it was best to just remove the test. The logic for `mkDots` is now so trivial that any test would basically just be a reimplementation of it 😛 

While I was at it I got rid of some unused imports in `Shorthands.hs`


madjar:
I think there is a line about `ShorthandTests` in the cabal file.

PR #92, opened 2018-03-01
Use environment in evaluation of function definitions

madjar:


madjar:
@jwiegley Would you prefer me to merge that kind of things directly, or would you prefer that I wait for another pair of eyes?

jwiegley:
This sort of thing you can just push directly, especially if it comes with a test.  Thanks!

PR #95, opened 2018-03-09
Add support for variadic parameters

madjar:
Everything was already in place, but not used for variadic parameters.

On the other hand fixed parameters sets are currently bugged, as they don't
reject extra arguments (they act like variadic parameters set). This commit adds
a TODO mentionning that.


jwiegley:
Thanks!

PR #97, opened 2018-03-17
Eval in main

madjar:
This PR changes main to also evaluate and print the value of the passed expression.

This would make it convenient for me when experimenting around evaluation, but I'm not sure if anybody depends on the current behaviour, so I'd like some opinion before merging.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -4,21 +4,30 @@ module Main where
 import Nix.Parser
 import Nix.Pretty
 import Nix.Expr
+import Nix.Eval (evalExpr)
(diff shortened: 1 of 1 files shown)


domenkozar:
This breaks backwards compatibility if someone used hnix as a pretty printer. But one can argue that was never intended to be so :) Maybe `hnix-eval`?

jwiegley:
Yes, I'd rather have it eval based on an option being used.

madjar:
@jwiegley Thanks for implementing this!
Closing the PR.

PR #134, opened 2018-04-04
Make thunks monadic values

Ericson2314:
Before, we just enforced that with the `pure $!` during elimination. Now
we store them as a monadic value with `pure $!` or similar effectively
done already. Effectively, instead of encapsulation on the construction
and elimination side, we only have encapsulation on the construction
side.

In both the old and the new cases, the invariant is that forcing a thunk
must be done in the Monad needed to run the thunk lest it not be cached.

This shouldn't change the correctness of the existing code, at least as
measured by the test suite. But hopefully it makes writing code going
forward easier.

### DIFF
--- a/Nix/Thunk.hs
+++ b/Nix/Thunk.hs
@@ -2,6 +2,9 @@
 
 module Nix.Thunk where
 
+import           Control.Exception.Base (evaluate)
+import           System.IO.Unsafe (unsafeInterleaveIO)
+
 -- | Rather than encoding laziness ourselves, using a datatype containing
 --   deferred actions or computed values in an IORef, we can leverage
 --   Haskell's own laziness by deferring actions until forced. For monads that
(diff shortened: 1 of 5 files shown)


Nix/Thunk.hs
```diff
@@ -12,19 +15,26 @@ module Nix.Thunk where
...
+--
 --   Many thanks to Theo Giannakopolous (@tgiannak) for clarifying this
 --   situation, and suggesting the use of MonadInterleave.
 
 class MonadInterleave m where
-    unsafeInterleave :: m a -> m a
+    unsafeInterleave :: m a -> m (m a)
 
-type Thunk (m :: * -> *) v = v
+type Thunk m v = m v
```
jwiegley, line 29: Should we make this `Either v (m v)`, so `valueRef` can store pure values?

PR #146, opened 2018-04-06
Implement __overrides

### LINKED ISSUE #123: Finish __overrides
See https://github.com/NixOS/nix/blob/master/src/libexpr/eval.cc#L820-L827

Fixes eval-okay-overrides

kmicklas:
This fixes #123.

### DIFF
--- a/Nix/Eval.hs
+++ b/Nix/Eval.hs
@@ -210,7 +210,7 @@ eval (NSet binds) = do
 
 eval (NRecSet binds) = do
     traceM "NRecSet..1"
-    s <- evalBinds True True binds
+    s <- evalBindsWithAlter (attrSetAlterWithOverride force) True True binds
     traceM $ "NRecSet..2: s = " ++ show (() <$ s)
     return $ NVSet s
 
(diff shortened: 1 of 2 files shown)


jwiegley:
Thanks for the effort, @kmicklas, but I think I'm going to go with what we have now, since it localizes this feature to just where binding occurs, and thus neither the core evaluator or `Lint.hs` has to change at all.  The whole of the code is now:

```
    buildResult . concat <=< mapM go . moveOverridesLast
  where
    moveOverridesLast = (\(x, y) -> y ++ x) .
        partition (\case NamedVar [StaticKey "__overrides"] _ -> True
                         _ -> False)

    go :: Binding (m v) -> m [([Text], m v)]
    go (NamedVar [StaticKey "__overrides"] finalValue) =
        finalValue >>= projectSet >>= \case
            Just o' -> return $ map (first (:[])) $
                fmap (fmap (\x -> forceThunk (coerce x) pure))
                           (M.toList o')
            x -> throwError $ "__overrides must be a set, but saw: "
                        ++ show (void x)
```

This way, if it ever does become deprecated, we just need to drop the call to `moveOverridesLast`, and the case that matches against `"__overrides"`, and nothing else would need to change.

kmicklas:
@jwiegley I noticed your code doesn't handle `inherit __overrides`.

jwiegley:
@kmicklas Oh, thank you for pointing that out!

PR #150, opened 2018-04-06
Replace pcre-light with regex-tdfa

mightybyte:
The regex-tdfa package implements POSIX extended regular expressions, which are what Nix uses.

### DIFF
--- a/Nix/Builtins.hs
+++ b/Nix/Builtins.hs
@@ -14,7 +14,7 @@
 {-# OPTIONS_GHC -Wno-missing-signatures #-}
 {-# OPTIONS_GHC -fno-warn-name-shadowing #-}
 
-module Nix.Builtins where --(MonadBuiltins, baseEnv) where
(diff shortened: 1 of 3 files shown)


Nix/Builtins.hs
```diff
@@ -394,12 +395,12 @@ match_ pat str = force pat $ \pat' -> force str $ \str' ->
         -- jww (2018-04-05): We should create a fundamental type for compiled
         -- regular expressions if it turns out they get used often.
         (NVStr p _, NVStr s _) -> return $ NVList $
-            let re = compile (encodeUtf8 p <> "$") []
-            in case match re (encodeUtf8 s) [exec_anchored] of
-                Nothing -> []
-                Just s  ->
+            let re = makeRegex (encodeUtf8 p) :: Regex
+            in case matchOnceText re (encodeUtf8 s) of
```
mightybyte, line 399: I suspect that matchOnceText will be more robust than appending a dollar sign to the regular expression.

PR #151, opened 2018-04-06
Test to catch a bug in builtins.match

mightybyte:
Also added infrastructure for comparing against current nix behavior,
not just current nix tests.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -35,6 +35,7 @@ let
         sha256 = "0l0g6ns5bcrcaij0wbdgc04qyl9h0vk1kx9lkzdkwj9v51l26azm";
       };
     };
+    modifier = drv: pkgs.haskell.lib.overrideCabal drv (old: { testHaskellDepends = old.testHaskellDepends ++ [pkgs.nix]; });
   };
 
   variant = if doBenchmark
--- a/default.nix
+++ b/default.nix
@@ -1,7 +1,32 @@
-{ pkgs ? import <nixpkgs> {} }: pkgs.haskellPackages.developPackage {
+let
+  hostPkgs = import <nixpkgs> {};
+  pinnedPkgs = hostPkgs.fetchFromGitHub {
+    owner = "NixOS";
+    repo = "nixpkgs-channels";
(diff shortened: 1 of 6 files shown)


default.nix
```diff
@@ -1,32 +1,7 @@
...
-  haskellPackages = let hpkgs = pkgs.haskell.packages.${compiler}; in
-    hpkgs // {
-      mkDerivation = args: hpkgs.mkDerivation (args // {
-        enableLibraryProfiling = doProfiling;
-        enableExecutableProfiling = doProfiling;
-      });
-    };
-
-  pkg = haskellPackages.developPackage {
+{ pkgs ? import <nixpkgs> {} }: pkgs.haskellPackages.developPackage {
```
jwiegley: This change loses the pinning, and the ability to enable profiling. DId you overwrite default.nix with something you intended only to use locally?

tests/EvalTests.hs
```diff
@@ -71,6 +72,8 @@ case_function_recursive_sets =
 case_nested_with =
     constantEqualStr "2" "with { x = 1; }; with { x = 2; }; x"
 
+case_match_failure_null = assertEvalMatchesNix "builtins.match \"ab\" \"abc\""
```
jwiegley, line 75: This is a great idea!

default.nix
mightybyte: Ack!  Yes.  Your `default.nix` file does not work for me.  Let me change it back.

jwiegley:
Before we install this, we have to decide what to do about Travis. Can we install Nix there?  Do you know @ryantrinkle?

mightybyte:
Yeah, we're getting this CI failure:

```
nix-instantiate: readCreateProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
```

I knew this was a possibility.  I was thinking we should probably add `nix` to `testHaskellDepends`.

jwiegley:
You can `nix` to `testHaskellDepends` in your PR, and Travis will give it a try.

jwiegley:
Merged manually.

PR #153, opened 2018-04-08
Change the nixpkgs pin

mightybyte:
The existing commit fails to build on my machine with the error:

```
api.c:22:10: fatal error: 'endian.h' file not found
#include <endian.h>
         ^~~~~~~~~~
1 error generated.
make[3]: *** [Makefile:693: libseccomp_la-api.lo] Error 1
make[2]: *** [Makefile:918: all-recursive] Error 1
make[1]: *** [Makefile:498: all-recursive] Error 1
make: *** [Makefile:407: all] Error 2
builder for '/nix/store/y77qz7z2s4skis94n82ji10lfiqcn0mv-libseccomp-2.3.3.drv' failed with exit code 2
```

This PR fixes the problem.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -2,9 +2,9 @@ let
   hostPkgs = import <nixpkgs> {};
   pinnedPkgs = hostPkgs.fetchFromGitHub {
     owner = "NixOS";
-    repo = "nixpkgs-channels";
-    rev = "ee28e35ba37ab285fc29e4a09f26235ffe4123e2";
-    sha256 = "0a6xrqjj2ihkz1bizhy5r843n38xgimzw5s2mfc42kk2rgc95gw5";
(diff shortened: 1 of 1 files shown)


jwiegley:
I'd rather not move away from the unstable channel just yet.

The fix to the problem you're encountering for the moment is to set `allowBroken = false` in your Nix configuration.

ryantrinkle:
Maybe we can add allowBroken=true; in our import of nixpkgs, just so it
doesn't depend on the user's config?

mightybyte:
The commit I'm changing to is on the unstable channel.  It's just an older commit that doesn't have the failure that I'm seeing.  The `nixpkgs` repo is the same as `nixpkgs-channels` in this case.

PR #154, opened 2018-04-08
Add builtins.split

mightybyte:


jwiegley:
I'd like to avoid a dependency on `directory`, since we already depend on `unix`.

jwiegley:
Actually, never mind that, they don't have an implementation of `canonicalizePath`

jwiegley:
@mightybyte Is this close to being ready?

mightybyte:
The split implementation is ready but it's still not passing yet because there's one test that includes an indented string and that's not working yet.

jwiegley:
Ah, since there's already an issue for that, we shouldn't hold up your split work. The scope of the PR just needs to be narrowed down to just the changes needed for split.

mightybyte:
Ok, then this should be ready.  It still won't pass the test case unless you comment out the one with the indented text.

jwiegley:
Merged by hand.

PR #174, opened 2018-04-13
Replace `cabal` with `runhaskell Setup.hs`

vaibhavsagar:
This mimics what `nixpkgs` does and works even if the user doesn't have `cabal` installed globally.


vaibhavsagar:
Thanks!

PR #182, opened 2018-04-17
use fromValue in more places in builtins

hSloan:
Do this to avoid pattern matching and throwing errors explicitly. 

Thanks for allowing me to contrib ^_^

### DIFF
--- a/src/Nix/Builtins.hs
+++ b/src/Nix/Builtins.hs
@@ -285,11 +285,7 @@ anyM p (x:xs)   = do
              else anyM p xs
 
 any_ :: MonadBuiltins e m => NThunk m -> NThunk m -> m (NValue m)
-any_ pred = flip force $ \case
-    NVList l ->
-        toNix =<< anyM fromNix =<< mapM (call1 pred) l
-    arg -> throwError $ "builtins.any takes a list as second argument, not a "
-              ++ show arg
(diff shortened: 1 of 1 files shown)


src/Nix/Builtins.hs
```diff
@@ -664,8 +649,8 @@ getEnv_ = fromNix >=> \s -> do
         Just v  -> Text.pack v
 
 sort_ :: MonadBuiltins e m => NThunk m -> NThunk m -> m (NValue m)
-sort_ comparator list = force list $ \case
-    NVList l -> NVList <$> sortByM cmp l
+sort_ comparator list = force list $ fromValue >=> \l ->
+    NVList <$> sortByM cmp l
```
jwiegley, line 653: you can use `toNix =<<` instead of `NVList <$>` in places like this.

jwiegley:
Thank you!

How is `genericClosure` treating you? :)

jwiegley:
@hSloan Oh, please hold off on further `Builtins.hs` changes like this. I'm working on a rather large refactoring within that file right now. You'll see the push when it comes through.

jwiegley:
I'm all done.

hSloan:
@jwiegley Will do, I'm planning on finishing off the `genericClosure` ticket next.

Issue #158, opened 2018-04-09
Write a QuickCheck generator for Nix expression trees


jwiegley:
There should actually be a way of using GHC Generics to synthesize AST generators.

But however we do it, since the parser and pretty should at least form an adjunction (it could become an isomorphism if we find a way to embed comments and whitespace), then this property should hold:

    parse . pretty ~~ id

It's not quite equal to identity, since the position information will differ, but it should be equal up to that information.

jwiegley:
It looks like `generic-random` will be good for this: https://hackage.haskell.org/package/generic-random

PR #229, opened 2018-04-27
Add QuickCheck test for pretty printer and parser

mpickering:
I have not analysed any of the failures or attempted to refine the
generator.

This would be an easy place for a new contributor to start by refining the generator and looking at why the tests are failing.

See #158

### DIFF
--- a/hnix.cabal
+++ b/hnix.cabal
@@ -2,7 +2,7 @@
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: b50a1c4cf11872cb711f0ed200482f16620ea7754fe7bf8cf8892413d0231746
+-- hash: ee27a667ebb4b2c1f121147d9491a677c0d0979c0d6547120ded75758329207a
 
 name:           hnix
 version:        0.5.0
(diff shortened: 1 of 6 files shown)


jwiegley:
The test is currently failing...

jwiegley:
@mpickering Can you fix the broken test so I can merge this wonderful work? Thanks!

mpickering:
Fixing the broken test involves fixing the pretty printer and parser which I don't have time to commit to right now.

jwiegley:
@mpickering Ah, understood, how much time do you estimate it to be, and what difficulty?

mpickering:
I didn't begin to look yet why it is failing. Perhaps at the weekend I will get the chance.

mpickering:
@jwiegley The test passes now with some refinement and two bug fixes.

jwiegley:
@mpickering I've had to disable these tests for the time being, until we've run through a lot more examples. The very first run on Travis broke, and I'm afraid that having it enabled will prevent CI from working until we've ironed out some of these issues.

The error can be seen with:

    hnix-tests --quickcheck-replay=678297

jwiegley:
@mpickering I've converted your tests to use Hedgehog rather than QuickCheck, since it gives us better shrinking for free.

PR #256, opened 2018-04-28
add div

wavewave:
div function is added.


wavewave:
we will need to add a test for this.

wavewave:
Checked with hnix. previously,
```
$ dist/build/hnix/hnix --eval --expr 'builtins.div 6 2'
hnix: Could not look up attribute div in "an attr set"
```
Now
```
$ dist/build/hnix/hnix --eval --expr 'builtins.div 6 2'
3
```
:-)

jwiegley:
The only problem is that this needs to work with floating point numbers as well. I'll take the PR as is, and add that support.

PR #263, opened 2018-04-28
add TopLevel placeHolder to builtins module

gdevanla:
Add the function placeHolder that hashes the output. Edit: a test has been added as well.

### DIFF
--- a/src/Nix/Builtins.hs
+++ b/src/Nix/Builtins.hs
@@ -267,7 +267,7 @@ unsafeGetAttrPos x y = x >>= \x' -> y >>= \y' -> case (x', y') of
                  ++ show (x, y)
 
 -- This function is a bit special in that it doesn't care about the contents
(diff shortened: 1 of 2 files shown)


shlevy:
@gdevanla Does the nix testsuite have a test of this?

src/Nix/Builtins.hs
```diff
@@ -167,6 +167,7 @@ builtinsList = sequence [
     , add  Normal   "parseDrvName"               parseDrvName
     , add2 Normal   "partition"                  partition_
     , add  Normal   "pathExists"                 pathExists_
+    , add' TopLevel   "placeHolder"                placeHolder
```
jwiegley: `placeholder` is all lowercase in the Nix C++ sources.

jwiegley (changes requested):

PR #265, opened 2018-04-28
getURL is introduced and implemented via http-client

wavewave:
For implementing various network operation, getURL is introduced as a method for `MonadEffects`.
I use http-client and http-client-tls for fetching the content from a designated URL.
Since we do not have a full implementation of hnix store yet, it just throws an error after downloading for now.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -36,8 +36,6 @@ in haskellPackages.developPackage {
       [
         pkgs.nix
         haskellPackages.hpack
-        haskellPackages.http-client
-        haskellPackages.http-client-tls
         # haskellPackages.cabal-install
       ];
 
--- a/default.nix
+++ b/default.nix
@@ -36,6 +36,7 @@ in haskellPackages.developPackage {
       [
         pkgs.nix
         haskellPackages.hpack
+        haskellPackages.HTTP
         # haskellPackages.cabal-install
       ];
 
--- a/default.nix
+++ b/default.nix
@@ -36,7 +36,8 @@ in haskellPackages.developPackage {
       [
         pkgs.nix
         haskellPackages.hpack
-        haskellPackages.HTTP
+        haskellPackages.http-client
+        haskellPackages.http-client-tls
         # haskellPackages.cabal-install
       ];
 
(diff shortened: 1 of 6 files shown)


default.nix
```diff
@@ -36,6 +36,8 @@ in haskellPackages.developPackage {
       [
         pkgs.nix
         haskellPackages.hpack
+        haskellPackages.http-client
```
shlevy: This shouldn't be necessary, `developPackage` should pick it up for you

shlevy (dismissed):

shlevy:
@imalsogreg Can you take a look? You were doing something here right?

default.nix
jwiegley: Right, we only need things here which are not referenced in `package.yaml`.

jwiegley (changes requested):
All looks good to me, after dropping that one line from default.nix.

hnix.cabal
```diff
@@ -95,6 +95,9 @@ library
     , exceptions
     , filepath
     , hashable
+    , http-types
```
jwiegley: We don't edit the `hnix.cabal` directly, but edit `package.yaml` and run `hpack`. There's a test that should have failed for you because of this...

jwiegley (changes requested):

wavewave:
thanks for the comments. changed!

package.yaml
```diff
@@ -68,6 +68,9 @@ library:
     - deriving-compat           >= 0.3 && < 0.5
     - directory
     - hashable
+    - http-types
+    - http-client
+    - http-client-tls
```
jwiegley, line 78: So, now that package.yaml is updated, which is correct, we also need to run hpack and commit the .cabal file so that people’s nix-shell will see it and for the Travis build.

jwiegley (changes requested):
Just one last bit needed...

package.yaml
wavewave: now i did it. thanks.

jwiegley (dismissed):

jwiegley:
This is being merged manually.

PR #269, opened 2018-04-29
Bultins.currentTime

hSloan:


shlevy (changes requested):
Discussed one on one, but needs to only call `getCurrentTime` once when evaluation state is initialized.

src/Nix/Builtins.hs
```diff
@@ -880,6 +881,11 @@ currentSystem = do
   arch <- getCurrentSystemArch
   return $ nvStr (arch <> "-" <> os) mempty
 
+currentTime :: (MonadEffects m, MonadNix e m) => m (NValue m)
```
mightybyte: You don't need the MonadEffects constraint here.  It is implied by MonadNix.

src/Nix/Builtins.hs
```diff
@@ -880,6 +881,11 @@ currentSystem = do
...
+currentTime = do
+  t <- getPosixTime
+  return . nvConstant . NInt $ fromIntegral $ fromEnum $ t
```
mightybyte: You don't need the last dollar sign here.

src/Nix/Exec.hs
```diff
@@ -570,10 +573,10 @@ instance (MonadFix m, MonadCatch m, MonadThrow m, MonadIO m,
         const $ toNix (0 :: Integer)
 #endif
 
-runLazyM :: Options -> MonadIO m => Lazy m a -> m a
-runLazyM opts = (`evalStateT` M.empty)
-              . (`runReaderT` newContext opts)
-              . runLazy
+runLazyM :: Options -> POSIXTime -> MonadIO m => Lazy m a -> m a
```
mightybyte: This change is probably going to cause errors in the test suite.  I don't see any other places in this PR where you make changes to code that calls runLazyM.

mightybyte (dismissed):

src/Nix/Exec.hs
jwiegley: I’d recommend making the current time an option, which gives you the ability to set it on the command line for testing purposes. Then you won’t need to add a new entry to the Context (even though I know I suggested it before).

Also, I’m more familiar with seeing UTCTime used in Haskell code than POSIXTime.

mightybyte: I probably led him astray there.  I too am more used to working with UTCTime, but the nix builtin wants POSIX time, so using Haskell's `POSIXTime` type seemed the most natural to me.  Your call.

src/Nix/Effects.hs
```diff
@@ -24,6 +25,8 @@ class MonadFile m => MonadEffects m where
     getCurrentSystemOS :: m Text
     getCurrentSystemArch :: m Text
 
+    getPosixTime :: m (POSIXTime)
```
shlevy: We're not using this effect anymore right?

hSloan: @shlevy I'm currently using this effect in `Builtins.hs`. Should I not be using that there anymore?

src/Nix/Exec.hs
hSloan: I'll make changes this evening. Thanks!

src/Nix/Effects.hs
hSloan: @shlevy Check out my latest changes when you get the chance. I will squash and rebase on your approval.

src/Nix/Builtins.hs
```diff
@@ -895,6 +899,12 @@ currentSystem = do
   arch <- getCurrentSystemArch
   return $ nvStr (arch <> "-" <> os) mempty
 
+currentTime :: MonadNix e m => m (NValue m)
+currentTime = do
+  opts :: NOpts.Options <- asks (view hasLens)
```
shlevy, line 933: @hSloan This looks right to me, but just to check @jwiegley is this the right way to use classy lenses with monad reader?

src/Nix/Builtins.hs
```diff
@@ -895,6 +899,12 @@ currentSystem = do
...
+  let t = fromMaybe (Time.posixSecondsToUTCTime 0) $ NOpts.currentTime opts
```
shlevy: Hmm I think we should error out if the current time is not initialized here. I'm not sure a `Maybe` is even the right interface.

src/Nix/Effects.hs
```diff
@@ -24,6 +25,8 @@ class MonadFile m => MonadEffects m where
     getCurrentSystemOS :: m Text
     getCurrentSystemArch :: m Text
 
+    getCurrentTime :: m (UTCTime)
```
shlevy: Now this isn't used right?

src/Nix/Exec.hs
```diff
@@ -572,8 +575,8 @@ instance (MonadFix m, MonadCatch m, MonadThrow m, MonadIO m,
 
 runLazyM :: Options -> MonadIO m => Lazy m a -> m a
 runLazyM opts = (`evalStateT` M.empty)
-              . (`runReaderT` newContext opts)
-              . runLazy
+    . (`runReaderT` newContext opts)
```
shlevy, line 620: This is no longer changed right?

src/Nix/Lint.hs
```diff
@@ -407,12 +407,12 @@ instance MonadThrow (Lint s) where
     throwM e = Lint $ ReaderT $ \_ -> unsafeIOToST $ throw e
 
 runLintM :: Options -> Lint s a -> ST s a
-runLintM opts = flip runReaderT (newContext opts) . runLint
+runLintM opts = flip runReaderT (newContext opts ) . runLint
```
shlevy: Cleanup

src/Nix/Lint.hs
```diff
@@ -407,12 +407,12 @@ instance MonadThrow (Lint s) where
...
 
 symbolicBaseEnv :: Monad m => m (Scopes m (SThunk m))
 symbolicBaseEnv = return emptyScopes
 
-lint :: Options -> NExprLoc -> ST s (Symbolic (Lint s))
+lint :: Options -> NExprLoc ->  ST s (Symbolic (Lint s))
```
shlevy: Cleanup

src/Nix/Options.hs
```diff
@@ -184,6 +188,9 @@ nixOptions = Options
         (   short 'f'
          <> long "file"
          <> help "Parse all of the files given in FILE; - means stdin"))
+    <*> (fmap (decode =<<) $ optional $ strOption
+        (   long "setTime"
+         <> help "Set current time for testing purposes"))
```
shlevy: Maybe `nixOptions` should take as an argument the actual current time, and use that by default if `setTime` isn't set?

Also I think `--current-time` may be a better flag than `--setTime`

tests/TestCommon.hs
```diff
@@ -33,7 +33,7 @@ hnixEvalText opts src = case parseNixText src of
     Failure err        ->
         error $ "Parsing failed for expressien `"
             ++ unpack src ++ "`.\n" ++ show err
-    Success expr ->
+    Success expr -> do
```
shlevy: Cleanup

shlevy (dismissed):
Are we actually setting the current time anywhere currently, if it's not on the command line?

src/Nix/Builtins.hs
```diff
@@ -62,6 +64,7 @@ import           Nix.Expr.Types
 import           Nix.Expr.Types.Annotated
 import           Nix.Frames
 import           Nix.Normal
+import qualified Nix.Options as NOpts
```
jwiegley, line 67: I don't use qualified names too much when the module doesn't contain overriding definitions. So you'll see it on HashMap, for example, but rarely on our own modules.

src/Nix/Builtins.hs
jwiegley: Yep, this is the right way!

src/Nix/Builtins.hs
jwiegley: The current time should always be set, and yes we should use `throwError` if we get a `Nothing` result here.

src/Nix/Effects.hs
jwiegley: Right, we don't need live querying of the time anymore.

src/Nix/Exec.hs
jwiegley: Looks like this is just a whitespace change, and can be dropped.

src/Nix/Lint.hs
jwiegley: At least I'm curious as to how that new space got there... :)

src/Nix/Options.hs
```diff
@@ -35,6 +37,7 @@ data Options = Options
     , arg          :: [(Text, Text)]
     , argstr       :: [(Text, Text)]
     , fromFile     :: Maybe FilePath
+    , currentTime  :: Maybe UTCTime
```
jwiegley, line 42: Ah, I see, it's a `Maybe` here so that the user doesn't have to specify it on the command-line.  That's the right type, we just have to make sure that if anyone ever uses the current time, it's an error if it was never set.

jwiegley: Actually, never mind that, we have it be just `UTCTime`, and then have both `defaultOptions` and `nixOptions` take a time value to use as the default if it's not specified.

src/Nix/Options.hs
jwiegley: `--now` is also a good option name.  Also see my comment above.

jwiegley (changes requested):

src/Nix/Lint.hs
hSloan: @jwiegley Sorry, I left it behind while undoing the addition of currentTime to Nix.Context

src/Nix/Builtins.hs
hSloan: @jwiegley Without importing qualified I get the following error: 
```
src/Nix/Builtins.hs:128:50: error:
    Ambiguous occurrence ‘currentTime’
    It could refer to either ‘Nix.Options.currentTime’,
                             imported from ‘Nix.Options’ at src/Nix/Builtins.hs:67:1-28
                             (and originally defined at src/Nix/Options.hs:42:7-17)
                          or ‘Nix.Builtins.currentTime’,
                             defined at src/Nix/Builtins.hs:906:1
```
Would you like me to rearrange some naming conventions?

jwiegley: Yeah, in cases like these, the effect should be named `currentTime`, and the function implementing the builtin should be named `currentTime_`.

jwiegley (approved):

jwiegley:
This is being merged manually.

Issue #234, opened 2018-04-27
Implement builtins.hasContext


jwiegley:

picnoir:
Looks like https://github.com/haskell-nix/hnix/pull/270 fixed this, I think we can close this issue.

PR #270, opened 2018-04-29
Add builtins.hasContext

rpglover64:
#234 

I would like to write a test for this, but I can't figure out how to generate a value which should have a context.

### DIFF
--- a/src/Nix/Builtins.hs
+++ b/src/Nix/Builtins.hs
@@ -261,9 +261,7 @@ attrsetGet k s = case M.lookup k s of
 
 hasContext :: MonadNix e m => m (NValue m) -> m (NValue m)
 hasContext x = fromValue @(Text, DList Text) >>= \(_, ctx) ->
-        toNix . not . null $ appEndo ctx []
-    x -> throwError $ ErrorCall $ "Invalid type for builtin.hasContext: "
-                 ++ show x
+    toNix . not . null $ appEndo ctx []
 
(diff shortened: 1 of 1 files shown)


rpglover64:
Does not work properly.

rpglover64:
To be fair, I'm not sure that the context is being constructed properly the code that checks it is pretty simple. Here's the worrying example:

```
[nix-shell:~/Documents/Haskell/hnix]$ ./dist/build/hnix/hnix --eval --expr 'let lib = (import <nixpkgs/lib>); pkg = import <nixpkgs> {}; in builtins.hasContext (baseNameOf "${./data/nix/tests/lang/eval-okay-context.nix}") '
false

[nix-shell:~/Documents/Haskell/hnix]$ nix-instantiate --eval --expr 'let lib = (import <nixpkgs/lib>); pkg = import <nixpkgs> {}; in builtins.hasContext (baseNameOf "${./data/nix/tests/lang/eval-okay-context.nix}") '
true
```

However, when I put in some debug-trace statements, the string passed to `builtins.hasContext` seems to have an empty context.

Reopening since the problem appears to be elsewhere.

shlevy:
Hoping to work with @gbwey9 on fixing up context handling today.

shlevy:
(the logic here seems right to me though)

jwiegley:
@shlevy Rather than create a new label to associate this issue with a "project", let's try creating a GitHub project! I've never used those before, and would love to give it a try.

jwiegley:
Merged manually.

PR #275, opened 2018-04-29
fix bug that made genEvalCompareTests not run any eval-compare tests

elaforge:


mightybyte (approved):
Nice catch!  Maybe we need some kind of a check in place to catch if tests get accidentally disabled in the future.

elaforge:
It's hard to assert that tests are run.  My best stab at it is that I use an automatic discovery mechanism that is simple and inflexible and never change it.  I guess in this case that just means don't add that *.nix filter in the first place :)

Issue #255, opened 2018-04-28
Add/document string context propagation


imalsogreg:
@shlevy

shlevy:
Hoping to tackle this with @gbwey9 today

PR #289, opened 2018-05-03
add string context 255

gbwey9:
Hi, 

Here is my attempt at doing #255 to add string context. 

I am particularly unsure about this change that I made in Eval.hs in MonadEval  
-    evalString      :: NString (m v) -> m v
+    evalString      :: NixString -> m v

Anyway let me know what I need to fix.

Best,
Grant


shlevy:
Will take a deep look this weekend!

src/Nix/Builtins.hs
```diff
@@ -250,7 +251,7 @@ nixPath = fmap nvList $ flip foldNixPath [] $ \p mn rest ->
 
 toString :: MonadNix e m => m (NValue m) -> m (NValue m)
 toString str =
-    str >>= normalForm >>= valueText False >>= toNix @(Text, DList Text)
+    str >>= normalForm >>= valueText False >>= toNix @NixString
```
jwiegley: I really like using `NixString` to abstract this tuple, and now can't believe it hadn't been done until now. :)

src/Nix/Builtins.hs
```diff
@@ -41,6 +41,7 @@ import           Data.Coerce
 import           Data.Fix
 import           Data.Foldable (foldrM)
 import qualified Data.HashMap.Lazy as M
+import qualified Data.HashSet as HS
```
jwiegley: I'd prefer if the qualified name was just `S`, and so didn't encode anything about which *kind* of Set is being used, since we might change that in future.

src/Nix/Eval.hs
```diff
@@ -111,7 +112,9 @@ eval (NSym var) =
     lookupVar var >>= maybe (freeVariable var) (force ?? evaledSym var)
 
 eval (NConstant x)          = evalConstant x
-eval (NStr str)             = evalString str
+eval (NStr str)             = assembleString str >>= \case
+                                Nothing -> evalError @v $ ErrorCall ("failed to evaluate string" :: String)
```
jwiegley: The `:: String` qualification isn't needed here, because `String` is the only type that `ErrorCall` accepts.

src/Nix/Value.hs
```diff
@@ -43,6 +44,38 @@ import           Nix.Frames
 import           Nix.Scope
 import           Nix.Thunk
 import           Nix.Utils
+import           Data.Semigroup
+
+-- | A 'ContextFlavor' describes the sum of possible derivations for string contexts
+data ContextFlavor =
```
jwiegley: Good stuff, thought I'll have to defer to @shlevy as to its correctness.

src/Nix/Value.hs
```diff
@@ -43,6 +44,38 @@ import           Nix.Frames
 import           Nix.Scope
 import           Nix.Thunk
 import           Nix.Utils
+import           Data.Semigroup
+
+-- | A 'ContextFlavor' describes the sum of possible derivations for string contexts
+data ContextFlavor = 
+    DirectPath
+  | DerivationOutput
```
shlevy: `DerivationOutput` should take a `Text` (referring to the specific output requested, e.g. `"out"`)

src/Nix/Value.hs
```diff
@@ -43,6 +44,38 @@ import           Nix.Frames
...
+  | AllDerivationOutput
```
shlevy: Nit: `AllDerivationOutputs`

shlevy (changes requested):
I'm thinking we should move the definition of `NixString` to its own module and not export the constructor/field names. We really don't want things like the implicit dropping of context in the `fromValueMay` definition, for example.

Instead, we should have something like (names certainly subject to change) `stringNoContext` which converts a `NixString` to a `Maybe Text`, being `Nothing` if the `NixString` has context, `stringIntentionallyDropContext` which converts a `NixString` to a `Text`, and `stringWithContext` which gives you both the `Text` and the `HashSet StringContext`. We should make it hard to use a `NixString` without explicitly thinking about what we want to do with the context.

jwiegley:
@shlevy I very much agree on moving it to its own module. We need more abstraction of that sort in `hnix`.

src/Nix/Builtins.hs
mightybyte: Yeah, this is vastly more understandable.

gbwey9:
Yes, moving NixString to its own module is a great idea! I'll get on that and also the other changes suggested by John above.

shlevy:
@gbwey9 Thanks! I'll take another pass this weekend.

src/Nix/Builtins.hs
```diff
@@ -266,20 +267,27 @@ attrsetGet k s = case M.lookup k s of
...
-    toNix . not . null . (appEndo ?? []) . snd <=< fromValue @(Text, DList Text)
+    toNix . not . null . stringContextOnly  <=< fromValue
 
 getAttr :: MonadNix e m => m (NValue m) -> m (NValue m) -> m (NValue m)
 getAttr x y = x >>= \x' -> y >>= \y' -> case (x', y') of
-    (NVStr key _, NVSet aset _) -> attrsetGet key aset >>= force'
+    (NVStr ns, NVSet aset _) -> 
+       case stringNoContext ns of
+        Just key -> attrsetGet key aset >>= force'
+        Nothing -> throwError $ ErrorCall $ "Invalid NixString for builtin.getAttr: "
```
shlevy: We should probably have a separate error type for this use case (maybe even just make `stringNoContext` throw it?). This is the error message C++ Nix gives us:

> nix-repl> builtins.getAttr "${./.}"  {} 
> error: the string '/nix/store/a8bwd9g6dvhbcr5q6pgh84i80zc0zxfq-hnix-store' is not allowed to refer to a store path (such as '/nix/store/a8bwd9g6dvhbcr5q6pgh84i80zc0zxfq-hnix-store'), at (string):1:1

src/Nix/Builtins.hs
```diff
@@ -393,7 +401,7 @@ splitVersion s = case Text.uncons s of
 splitVersion_ :: MonadNix e m => m (NValue m) -> m (NValue m)
 splitVersion_ = fromValue >=> \s -> do
```
shlevy, line 471: Wait, does this imply there's a `fromValue` instance that gives you `Text`? We definitely don't want that!

src/Nix/Builtins.hs
```diff
@@ -666,7 +674,7 @@ toPath = fromValue @Path >=> toNix @Path
 pathExists_ :: MonadNix e m => m (NValue m) -> m (NValue m)
 pathExists_ path = path >>= \case
     NVPath p  -> toNix =<< pathExists p
-    NVStr s _ -> toNix =<< pathExists (Text.unpack s)
+    NVStr ns -> toNix =<< pathExists (Text.unpack (stringIntentionallyDropContext ns))
```
shlevy: `pathExists` should fail if the string has context.

src/Nix/Builtins.hs
```diff
@@ -745,7 +753,7 @@ lessThan ta tb = ta >>= \va -> tb >>= \vb -> do
             (NInt   a, NFloat b) -> pure $ fromInteger a < b
             (NFloat a, NFloat b) -> pure $ a < b
             _ -> badType
-        (NVStr a _, NVStr b _) -> pure $ a < b
+        (NVStr a, NVStr b) -> pure $ a < b
```
shlevy: In the C++ code, `lessThan` only compares the underlying strings, not the contexts... what does `Ord` for `NixString` do?

src/Nix/Builtins.hs
```diff
@@ -779,8 +787,8 @@ placeHolder = fromValue @Text >=> \_ -> do
 
 absolutePathFromValue :: MonadNix e m => NValue m -> m FilePath
 absolutePathFromValue = \case
-    NVStr pathText _ -> do
-        let path = Text.unpack pathText
+    NVStr ns -> do
+        let path = Text.unpack $ stringIntentionallyDropContext ns
```
shlevy: For now, we should fail if the string has context, but actually in the C++ code base this works through the same mechanism as import-from-derivation: If we do readFile or readDir on a path with conext, we build the derivations we need to build during evaluation and *then* read from the file/directory.

@jwiegley What's the best way to track that we need to change this once we support IFD?

src/Nix/Builtins.hs
```diff
@@ -828,7 +836,7 @@ fromJSON = fromValue >=> \encoded ->
 
 toXML_ :: MonadNix e m => m (NValue m) -> m (NValue m)
 toXML_ v = v >>= normalForm >>= \x ->
-    pure $ nvStr (Text.pack (toXML x)) mempty
+    pure $ nvStr $ makeNixStringWithoutContext $ Text.pack (toXML x)
```
shlevy: `toXML` should output a `NixString`, it may have some context depending on the value in question.

src/Nix/Builtins.hs
```diff
@@ -884,7 +892,7 @@ fetchTarball v = v >>= \case
  where
     go :: Maybe (NThunk m) -> NValue m -> m (NValue m)
     go msha = \case
-        NVStr uri _ -> fetch uri msha
+        NVStr ns -> fetch (stringIntentionallyDropContext ns) msha
```
shlevy: URL can't have any context.

src/Nix/Builtins.hs
```diff
@@ -920,7 +928,7 @@ fetchurl v = v >>= \case
  where
     go :: Maybe (NThunk m) -> NValue m -> m (NValue m)
     go _msha = \case
-        NVStr uri _ -> getURL uri -- msha
+        NVStr ns -> getURL (stringIntentionallyDropContext ns) -- msha
```
shlevy: URL can't have any context.

src/Nix/Eval.hs
```diff
@@ -112,7 +113,9 @@ eval (NSym var) =
     lookupVar var >>= maybe (freeVariable var) (force ?? evaledSym var)
 
 eval (NConstant x)          = evalConstant x
-eval (NStr str)             = evalString str
+eval (NStr str)             = assembleString str >>= \case
```
shlevy: Hmm, why did this change?

src/Nix/Exec.hs
```diff
@@ -178,13 +179,11 @@ instance MonadNix e m => MonadEval (NValue m) m where
...
-            scope <- currentScopes
-            span  <- currentPos
-            pure $ nvStrP (Provenance scope
-                           (NStr_ span (DoubleQuoted [Plain s]))) s c
-        Nothing -> nverr $ ErrorCall "Failed to assemble string"
+    evalString ns = do
+        scope <- currentScopes
+        span  <- currentPos
+        pure $ nvStrP (Provenance scope
+                           (NStr_ span (DoubleQuoted [Plain (stringIntentionallyDropContext ns)]))) ns
```
shlevy: Hmm it's hard to tell here what `evalString` is doing... @jwiegley What's this used for? We definitely shouldn't drop context in a generic function...

src/Nix/Exec.hs
```diff
@@ -428,7 +429,7 @@ coerceToString = \case
     NVConstant (NUri u)   -> pure $ show u
     NVConstant NNull      -> pure ""
 
-    NVStr t _ -> pure $ Text.unpack t
+    NVStr ns -> pure $ Text.unpack $ stringIntentionallyDropContext ns
```
shlevy: Not sure where `coerceToString` is being used, but if it's inspired by the function of the same name in the C++ codebase we definitely don't want to drop context here.

src/Nix/Normal.hs
```diff
@@ -86,13 +87,13 @@ valueText addPathsToStore = cata phi
     phi (NVPathF originalPath)
         | addPathsToStore = do
             storePath <- addPath originalPath
-            pure (Text.pack $ unStorePath storePath, mempty)
-        | otherwise = pure (Text.pack originalPath, mempty)
+            pure (makeNixStringWithoutContext $ Text.pack $ unStorePath storePath)
```
shlevy: We want to add the new path to the context here, not drop it.

src/Nix/Normal.hs
```diff
@@ -86,13 +87,13 @@ valueText addPathsToStore = cata phi
...
+        | otherwise = pure (makeNixStringWithoutContext (Text.pack originalPath))
     phi v@(NVBuiltinF _ _) = coercionFailed v
 
     coercionFailed v =
         throwError $ Coercion @m (valueType v) TString
 
 valueTextNoContext :: (Framed e m, MonadEffects m, Typeable m)
                    => Bool -> NValueNF m -> m Text
-valueTextNoContext addPathsToStore = fmap fst . valueText addPathsToStore
+valueTextNoContext addPathsToStore = fmap stringIntentionallyDropContext . valueText addPathsToStore
```
shlevy: What is this function for?

src/Nix/XML.hs
```diff
@@ -26,7 +27,7 @@ toXML = (.) ((++ "\n") .
         NNull    -> Element (unqual "null") [] [] Nothing
         NUri u   -> mkElem "uri" "value" (Text.unpack u)
 
-    NVStrF t _ -> mkElem "string" "value" (Text.unpack t)
+    NVStrF ns -> mkElem "string" "value" (Text.unpack $ stringIntentionallyDropContext ns)
```
shlevy: This ultimately needs to be changed to output a NixString, not a Text

shlevy (changes requested):
Sorry for the large number of changes... Let me know if you want to chat synchronously at some point to help figure out the best way forward, or if you just want to keep going addressing one by one.

src/Nix/Builtins.hs
jwiegley: A separate error type would be nice.

src/Nix/Builtins.hs
jwiegley: Why not? It's OK to drop context info you're not going; you just don't want a `toValue` that forgets it when constructing a new string.

src/Nix/Builtins.hs
jwiegley: Good catch.

src/Nix/Builtins.hs
jwiegley: Track it with a GitHub issue?

src/Nix/Exec.hs
jwiegley: `evalString` is recording provenance while building a string. Sounds like context needs to be preserved here.

src/Nix/Exec.hs
jwiegley: It's meant to become an exact replica of the C++ version.

src/Nix/Builtins.hs
shlevy: @jwiegley I think we want to make it hard to accidentally drop context. Some primops should fail on non-empty context, others need to do special work (e.g. IFD) on non-empty context, even if there's no new string being constructed.

shlevy:
@gbwey9 I was thinking it might be useful for me to write up an extensive document describing string context from the perspective of Nix language semantics, abstracted away from the specific implementation. Does that seem helpful?

gbwey9:
@shlevy Yes a document describing string context would be really helpful, thanks! What would you recommend for getting up to speed on Nix (or at least the stuff I need to know)? I am currently going through the Nix pills series.

shlevy:
@gbwey9 For the stuff you're working on, I definitely recommend Eelco's thesis! https://nixos.org/~eelco/pubs/phd-thesis.pdf

gbwey9:
@shlevy Thanks I'll take a look.

shlevy:
Hi @gbwey9! So sorry I dropped this! Are you still interested in working on it if I get that doc together for you?

gbwey9:
Hi @shlevy , I am still interested in working on this and would be really happy to get that doc from you.
However, is there some way to break up this pull request into smaller more manageable pieces?
I am not sure if this is possible otherwise, as nix string impacts so much of the code.

jwiegley:
@gbwey9 As a very first step, let's abstract out a `NixString` data type, and plumb that change throughout the code, and after that start adding context info.

gbwey9:
@jwiegley Great idea. I'll work on that.

shlevy:
OK, great! I'll pick up the documentation work then, and we can figure out the right path from there.

gbwey9:
As you suggested, I have plumbed Nix through the codebase. 
Let me know what corrections I need to make.
Best,
Grant

shlevy:
@gbwey9 A bunch of us are gathering this weekend (you're welcome to join if you're near NH or want to connect remotely!), we'll take another look at this then!

gbwey9:
@shlevy that sounds great. I can't make it in person but would be glad to connect remotely. 
I do appreciate any help you and the others can give me on this

jwiegley:
@gbwey9 Maybe we could phone you in by Skype?

gbwey9:
@jwiegley Sounds good. However, can i just call in? I don't yet have a Skype account but I can set one up if that will work better.

jwiegley:
@gbwey9 Yes, we could put you on speaker phone for a bit.

gbwey9:
@jwiegley Thanks.

shlevy:
@gbwey9 How should we contct you?

gbwey:
my number is 1-734-274-5299 or I can call you.

shlevy:
@gbwey texted you

gbwey:
Sorry I have no idea where that text went. but my email address is weyburne@gmail.com

jwiegley:
@gbwey @shlevy How is our progress on the string context stuff? Have we finished creating a new NixString type?

gbwey9:
@jwiegley I have been actively working on it this past week and will ask for some more help from Shea in order to finish it off.

src/Nix/NixString.hs
```diff
@@ -0,0 +1,68 @@
+{-# LANGUAGE DeriveGeneric #-}
+module Nix.NixString (
+    NixString
+  , hackyStringHasContext
+  , hackyStringIgnoreContextMaybe
+  , hackyStringIgnoreContext
+  , makeNixStringWithoutContext
```
shlevy: We should have a `hacky` variant of this and use it everywhere for now. While making a string without context is something we want to expose, we still need to go through all current string creation to ensure it ought to be contextless

src/Nix/NixString.hs
```diff
@@ -0,0 +1,68 @@
+{-# LANGUAGE DeriveGeneric #-}
+module Nix.NixString (
+    NixString
+  , hackyStringHasContext
```
shlevy: This one is actually non-hacky for free :)

src/Nix/Builtins.hs
shlevy: @gbwey9 FYI, this is OK for now but once you try to remove the `hacky` bits from the `FromValue` instance we'll definitely have to address this properly.

src/Nix/NixString.hs
```diff
@@ -0,0 +1,68 @@
...
+  , hackyStringIgnoreContextMaybe
+  , hackyStringIgnoreContext
+  , makeNixStringWithoutContext
+  , modifyNixContents
```
shlevy: I think I'd prefer a `hacky` version of this as well, most of these are probably correct as-is but some string modifications absolutely need to change context as well.

src/Nix/NixString.hs
```diff
@@ -0,0 +1,68 @@
...
+import           Data.Text (Text)
+import           GHC.Generics
+import           Data.Semigroup
+
+{-# WARNING hackyStringHasContext, hackyStringIgnoreContextMaybe, hackyStringIgnoreContext "This NixString function needs to be replaced" #-}
+
+-- | A 'ContextFlavor' describes the sum of possible derivations for string contexts
+data ContextFlavor = 
+    DirectPath
+  | DerivationOutput !Text
```
shlevy: Can we have some more haddocks on these constructors?

src/Nix/NixString.hs
```diff
@@ -0,0 +1,68 @@
...
+
+data NixString = NixString 
+  { nsContents :: !Text
+  , nsContext :: !(S.HashSet StringContext)
+  } deriving (Eq, Ord, Show, Generic) 
+
+instance Hashable NixString
+
+instance Semigroup NixString where
+  NixString s1 t1 <> NixString s2 t2 = NixString (s1 <> s2) (t1 <> t2)
```
shlevy: Let's leave this for now, but I am a tiny bit concerned that it might let context mishandling sneak in somewhere. Just to flag it in case we're seeing weird bugs in the future :joy:

shlevy (changes requested):
Looking great, much nicer! In addition to inline comments, a few top-level thoughts:

1. It seems like your editor marked all files executable and changed newline endings for the READMEs/LICENSEs. I'm wondering if that's some Windows setting, can you see if that can be fixed?
2. The commit history is quite messy here, once this is stabilized we should probably squash it down to something more useful to people looking at the logs

src/Nix.hs
```diff
@@ -11,6 +11,7 @@ module Nix (module Nix.Cache,
             module Nix.Render.Frame,
             module Nix.Normal,
             module Nix.Options,
+            module Nix.NixString,
```
jwiegley: I think, for consistency's sake, that while `NixString` is a fine name for the type, the module is better named just `Nix.String`. `Nix.NixString` seems redundant...

src/Nix/Builtins.hs
```diff
@@ -100,7 +101,7 @@ withNixContext mpath action = do
     opts :: Options <- asks (view hasLens)
     let i = value @(NValue m) @(NThunk m) @m $ nvList $
             map (value @(NValue m) @(NThunk m) @m
-                     . flip nvStr mempty . Text.pack) (include opts)
+                     . nvStr . hackyMakeNixStringWithoutContext . Text.pack) (include opts)
```
jwiegley, line 104: Does the function really need to be named `hackyMakeNixStringWithoutContext`? Is this some indication that the work is yet undone?

jwiegley (changes requested):
It's starting to look pretty good!

src/Nix/Builtins.hs
shlevy: Yes. We need to ensure we look at all places where we create strings and set the context appropriately. For some of those it will be correct to just make a string without context, of course, but currently *every* string creation in hnix is without context and that's definitely wrong.

gbwey9: @shlevy As you mentioned before, there are two more places where we use mconcat (Eval.hs) and mappend (Exec.hs) on NixStrings. Should we just get rid of the semigroup/monoid instance and replace those calls with NixString specific calls?

src/Nix/Exec.hs
```diff
@@ -336,7 +336,7 @@ execBinaryOp scope span op lval rarg = do
             _                  -> nverr $ ErrorCall $ unsupportedTypes lval rval
 
         (NVStr ls, NVStr rs) -> case op of
-            NPlus -> pure $ bin nvStrP (ls `mappend` rs)
+            NPlus -> pure $ bin nvStrP (ls `hackyStringMappend` rs)
```
jwiegley, line 339: Can you explain a bit more of the motivation behind this?

jwiegley (changes requested):

src/Nix/Exec.hs
gbwey9: It may be that some of the NixString function calls are incorrect, so the first step is to identify every instance. Then we can systematically evaluate each occurrence and then assign it a non-hacky* function from NixString. In other words, this first step is solely to mark these NixString calls.

src/Nix/Builtins.hs
shlevy: @gbwey9 Yeah, I think that makes the most sense. We can add it back in once we've cleaned everything up and found the right patterns for using context.

src/Nix/Exec.hs
jwiegley: As long as at some point in the future we go back to using `Monoid` if this thing turns out to be a monoid, I'm fine with interim measures.

gbwey9:
I just noticed that I missed the NixString change from Nix.NixString to Nix.String.
Also the travis build (strict) fails because of the deprecation warnings that are on the some of the NixString methods but builds otherwise. Should I remove those warnings from Nix.String?

gbwey9:
@shlevy @jwiegley Is there anything else I need to do to get this pull request through?
I am happy to work on this, so just let me know.

shlevy (approved):
Looks great! I'd prefer a squash before merge but I'll leave that up to @jwiegley

jwiegley:
I avoid squash, both at the dinner table and in Git. I'll merge once CI is happy.

jwiegley (approved):

gbwey9:
@jwiegley @shlevy  Thanks so much

PR #293, opened 2018-05-04
WIP making hnix build with GHCJS

mightybyte:
This is ready for review, but there are a couple outstanding items.

* Currently `builtins.derivation` is disabled with CPP for GHC < 8.2.  This was needed because at the moment GHCJS only works with GHC 8.0.
* The console REPL code has been removed because it cannot be built with GHCJS.  It will be moved into another repo.


default.nix
```diff
@@ -18,23 +18,29 @@
...
-        });
-      ghc-heap-view =
-        pkgs.haskell.lib.overrideCabal super.ghc-heap-view (attrs: {
-          enableLibraryProfiling    = false;
-          enableExecutableProfiling = false;
-        });
-    };
+    overrides = with pkgs.haskell.lib; self: super:
+      if compiler == "ghcjs" then {} else
+      rec {
```
jwiegley: `rec` isn't needed here, since there are no internal references to the set members.

main/Main.hs
```diff
@@ -1,146 +0,0 @@
-{-# LANGUAGE FlexibleContexts #-}
```
jwiegley: Removing the whole main driver from `hnix` into its own project will end up making things a real pain during active development, giving the way that Nix doesn't support incremental builds. What I'd rather do is just change `package.yaml` so that it doesn't build the main driver or the repl when a ghcjs build is requested. Until until the core has fully stabilized, (hopefully) a few months from now.

src/Nix/Exec.hs
```diff
@@ -440,11 +439,6 @@ instance MonadCatch m => MonadCatch (Lazy m) where
 instance MonadThrow m => MonadThrow (Lazy m) where
     throwM = Lazy . throwM
 
-instance MonadException m => MonadException (Lazy m) where
```
jwiegley: Rather than make this an orphan instance in another package, why don't we just `#if` it out when ghcjs is being used?

tests/NixLanguageTests.hs
```diff
@@ -111,24 +111,8 @@ assertEval _opts files =
         [".exp"] -> assertLangOk defaultOptions name
         [".exp.disabled"] -> return ()
         [".exp-disabled"] -> return ()
-        [".exp", ".flags"] -> do
```
jwiegley: I'd much rather we just `#ifdef` this out for ghcjs, than remove it. This change would mean the core library can't be tested against the Nix language tests anymore.

jwiegley (changes requested):

mightybyte:
Hmm, this CI failure doesn't seem like a legit failure.

jwiegley (dismissed):
This all looks great to me. @shlevy @ryantrinkle any thoughts?

jwiegley:
@mightybyte I've restarted the CI build. If it happens again, we'll need to track it down.

Also, do you know what we need to add to `.travis.yml` to ensure that we keep the GHCJS code building after this merge?

mightybyte:
Ugh...the tests pass, but it failed at the end again.  It looks like it's something to do with the haddocks and rebuilding the cabal file.

```
Running 1 test suites...
Test suite hnix-tests: RUNNING...
Test suite hnix-tests: PASS
Test suite logged to: dist/test/hnix-0.5.0-hnix-tests.log
1 of 1 test suites (1 of 1 test cases) passed.
haddockPhase
./hnix.cabal has been changed. Re-configuring with most recently used options.
If this fails, please run configure manually.
Configuring hnix-0.5.0...
Setup: Use of GHC's environment variable GHC_PACKAGE_PATH is incompatible with
Cabal. Use the flag --package-db to specify a package database (it can be used
multiple times).
```

jwiegley:
@mightybyte What could be changing the `.cabal` file after running the tests? I think we need to track this down before merging.

mightybyte:
Grrrr...the cabal file was out of sync with package.yaml.  I don't know about anyone else, but hpack here has been making things more painful for me, not less.

jwiegley:
@mightybyte Basically: never change the `.cabal`. Ever. You could also create a pre-commit hook that bombs out if your `.cabal` is newer than the `.yaml`.

Best of all, a fix to `developPackage` so that it's not slow in the absence of a `.cabal` would let us get rid of that file entirely from the repo.

mightybyte:
I don't think fixing `developPackage` would allow us to get rid of the `.cabal` file because once we put this on hackage the cabal file will be needed.

jwiegley:
So, I woke up today wondering about [Jupyter Notebook](http://jupyter.org/), and found from reading their documentation that it wouldn't be hard at all to write a Kernel to support the Nix language. You just have to talk to the REPL loop via ZeroMQ, it looks like.

This led me to wonder: Why exactly do we want Nix evaluation to happen *inside* the browser? The main advantage to this that I can think of is fully transparent introspection of all the data associated with an error message. However, since Emacs too is going to need such introspection, we're going to need to develop some kind of JSON-based error reporting anyway in order to allow fully drilling down.

So as I can see it right now, there are some pros and cons to supporting GHCJS:

  - Pro: You can inspect everything directly, in Haskell, without a wire protocol between.
  - Con: You can inspect everything directly, in Haskell, without a wire protocol between.  I list this also as a con because forcing us to use a wire protocol would dramatically improve that protocol for other clients.
  - Con: The code must be conditionalized for building in GHCJS, and for every new dependency we have to re-ask the question, "Will this work with GHCJS and how?"  Unless someone is very active at keeping the GHCJS support working (Hi @mightybyte), I could see it falling behind. Already today we've brought in several new dependencies for testing and convenience purposes.
  - Maybe con: The browser doesn't have access to the store, so unless we're going to always use a mock store, aren't GHCJS evaluations limited to pure invocations only? Is this valuable?
  - Pro: We don't know yet how others will want to use Nix as a tooling library; being able to execute in a Javascript environment gives us pretty much maximum flexibility.

jwiegley:
The requirement of having a `.cabal` in the sdist tarball is unrelated to whether or one we have a copy living under version control.

jwiegley:
Pinging @shlevy and @ryantrinkle for your opinions on the above musings about GHCJS.

shlevy:
I think maintaining ghcjs support currently seems like a lot of effort, but I also think Doug will keep it up if he commits to it. Personally all of the things I might want to exist based on an in-browser implementation would be equally or better served by a proper wire protocol, but I'm happy to leave it up to those who will actually implement those tools to decide which approach works best for them.

mightybyte:
If we do merge this, I definitely want to get GHCJS added to the travis script so it will be easy to maintain GHCJS buildability.  But I wanted to get this merged first to avoid continued merges.

jwiegley:
I'm certainly willing to let it in and see how it goes. I may mark the GHCJS builds as "allowed to fail", so that the build doesn't stay broken until Doug can patch up the GHCJS support. He'll just have to keep an eye on that build result.

jwiegley:
I've manually merged and reworked this changeset, so I'm going to push. Also, deleting the `minimize-deps` branch, since it just shows as broken in Travis.

PR #301, opened 2018-05-06
Make use of lens-family to provide some utility lenses/traversals

jwiegley:


jwiegley:
@mightybyte I'm wondering if these will work with GHCJS or not?

jwiegley:
Pull-request updated, HEAD is now 0b05ab07814aa3cd390a017a21fd649232e4081e

PR #302, opened 2018-05-06
Fix #280. unsafeGetAttrPos should return null.

### LINKED ISSUE #280: eval-okay-getattrpos-undefined
This appears to be a new test in `nix`:

```
Exception: unsafeGetAttrPos: field 'abort' does not exist in attr set: fromList []
```

BlairArchibald:
According to the expected test output, unsafeGetAttrPos should fail silently and
return null if it can't find a position for the field (e.g when looking up
builtins).


jwiegley:
There a simpler way now to do this kind of check, so I'll merge this and then make that change. Thanks!

PR #303, opened 2018-05-06
Distinguish between the three different types of string coercion

jwiegley:


jwiegley:
@shlevy @gbwey9 This code isn't quite working yet, but should be developed in tandom with the string context work, I think. @gbwey9 Do you want to take this code over from here?

gbwey9:
Yes, I'll get to work on this one.

shlevy:
Yeah, this is a natural consequence of getting string context right.

jwiegley:
This will be obviated by other work coming in soon from @shlevy.

PR #313, opened 2018-05-12
Another round of fixes for GHCJS

mightybyte:


package.yaml
```diff
@@ -165,6 +164,7 @@ executables:
           buildable: false
         else:
           buildable: true
+          dependencies: pretty-show
```
jwiegley, line 167: Why is this dependency moved here?  If the flag is true, the entire target will not be built.

package.yaml
```diff
@@ -188,13 +188,13 @@ tests:
...
       - megaparsec
       - tasty-quickcheck
-      - pretty-show
     when:
       - condition: "impl(ghcjs)"
         then:
           buildable: false
         else:
           buildable: true
+          dependencies: pretty-show
```
jwiegley, line 197: Same.

src/Nix/Exec.hs
```diff
@@ -753,7 +755,11 @@ addTracing k v = do
             opts :: Options <- asks (view hasLens)
             let rendered =
                     if verbose opts >= Chatty
+#ifdef MIN_VERSION_pretty_show
```
jwiegley, line 758: If Main.hs isn't being built for ghcjs, then none of these conditional statements are needed, unless some version of GHC needs them.

jwiegley (changes requested):

package.yaml
mightybyte: Apparently not.  I tried it the other way and it didn't work.  The `pretty-show` package does not build with GHC, so I had to move it here.  This might be because `cabal2nix` does not take build flags into account.

src/Nix/Exec.hs
mightybyte: This was the minimal change I could find to get it building.

package.yaml
jwiegley: It doesn't build with GHC, or GHCJS?

I'll use your PR to reproduce the build here. There has to be a way to get this working that's sane; moving dependencies under a `Buildable` guard feels like too much of a hack.

src/Nix/Exec.hs
jwiegley: OK, I'll give it a go later tonight and see what I can find, thanks.

package.yaml
mightybyte: GHCJS

src/Nix/Exec.hs
jwiegley: Hmm, it would seem that you're not using the version of `hpack` that's specified in `hnix`s `default.nix`, since all the Travis builds are failing due to the `.cabal` being out of sync again.

mightybyte: Oh, I probably forgot to re-run hpack.  I'll push an update.

package.yaml
jwiegley: @mightybyte I just tried merging this branch but undoing your `package.yaml`/`hnix.cabal` changes, and everything still built fine using `ghcjs.nix`. What do I need to do to see the error you ran into that motivated this change?

src/Nix/Exec.hs
jwiegley: Ah, I didn't realize this was in `Exec.hs`, not `Main.hs`, so these changes are good.

package.yaml
mightybyte: Weird, I don't know.  I could swear it wasn't building for me without that, but I'm not sure what would be different.  The only thing I can think of is that maybe I made the `hnix` code changes first and then subsequently made changes to the hnix-web-repl `default.nix` file that made those changes irrelevant.  I'll give it a try with the new hnix master.

PR #315, opened 2018-05-14
Abort on nix prior to 2.0

alexfmpe:


jwiegley (approved):
Sounds like a great idea to me.

default.nix
```diff
@@ -5,12 +5,14 @@
...
 , rev         ? "9d0b6b9dfc92a2704e2111aa836f5bdbf8c9ba42"
 , sha256      ? "096r7ylnwz4nshrfkh127dg8nhrcvgpr69l4xrdgy3kbq049r3nb"
-, nixpkgs     ? import (builtins.fetchTarball {
-    url = "https://github.com/NixOS/nixpkgs/archive/${rev}.tar.gz";
-    inherit sha256; }) {
-    config.allowUnfree = true;
-    config.allowBroken = false;
-  }
+, nixpkgs     ? if builtins.compareVersions builtins.nixVersion "2.0" < 0
+                then abort "hnix requires nix 2.0"
```
jwiegley, line 9: Shouldn't this check for "at least 2.0"? What happens when 2.1 comes out?

jwiegley (changes requested):

default.nix
alexfmpe: I believe this aborts when nixVersion is "smaller" than 2.0, so 2.1 should work fine.
The abort message itself could say "at least 2.0" though.

jwiegley: I made that change, thanks!

PR #316, opened 2018-05-14
Add HPACK_TESTS and an ALL_TESTS env variable

Profpatsch:
See PR description, `HPACK_TESTS` is used to be able to run the tests in nixpkgs without pulling in `hpack`, `ALL_TESTS` should be convenient.


Profpatsch:
Ah, uh, apparently the new remote is not a redirect, but a comptely new repo? Strange.

jwiegley (dismissed):
I like it.

Profpatsch:
Okay, rebased.

PR #320, opened 2018-05-24
Add back cabal-install to default.nix

alexfmpe:
I'm not sure if this is the proper fix, but I can't follow the README instructions since https://github.com/haskell-nix/hnix/commit/c9e1393391dd454b3ae9c49b489d8c4b97fe4617 removed cabal-install.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -93,7 +93,16 @@ in haskellPackages.developPackage {
         # .cabal file will be. Otherwise, Travis may error out claiming that
         # the cabal file needs to be updated because the result is different
         # that the version we committed to Git.
-        pkgs.haskell.packages.ghc822.hpack ];
+        pkgs.haskell.packages.ghc822.hpack
+
+        (let cabalInstallVersion = {
+               ghc802 = "1.24.0.2";
+               ghc822 = "2.0.0.1";
+               ghc842 = "2.2.0.0";
+             }; in
+         haskellPackages.callHackage "cabal-install"
+          cabalInstallVersion.${compiler} {})
+      ];
 
     enableLibraryProfiling    = doProfiling;
     enableExecutableProfiling = doProfiling;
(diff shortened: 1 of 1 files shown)


jwiegley:
I'm happy to merge this, once I figure out why CI doesn't like it.

picnoir:
We miss a cabal version entry for ghc 8.4.3.

I don't think I can push the diff to this pr:

```
diff --git a/default.nix b/default.nix
index a45da95..8030e54 100644
--- a/default.nix
+++ b/default.nix
@@ -79,6 +79,7 @@ in haskellPackages.developPackage {
             ghc802 = "1.24.0.2";
             ghc822 = "2.0.0.1";
             ghc842 = "2.2.0.0";
+            ghc843 = "2.2.0.0";
         }; in
         haskellPackages.callHackage "cabal-install"
         cabalInstallVersion.${compiler} {})
```

[Edit] Cabal 2.2.0.1 is not available from our nixpkgs revision, let's use the 2.2.0.0.

default.nix
```diff
@@ -75,7 +75,17 @@ in haskellPackages.developPackage {
...
-        pkgs.haskell.packages.ghc822.hpack ];
+        pkgs.haskell.packages.ghc822.hpack
+
+        (let cabalInstallVersion = {
+               ghc802 = "1.24.0.2";
+               ghc822 = "2.0.0.1";
+               ghc842 = "2.2.0.0";
+               ghc843 = "2.2.0.0";
+             }; in
+         haskellPackages.callHackage "cabal-install"
```
matthewbauer, line 86: I think it would be more straightforward to just do ```pkgs.haskell.packages.${compilerName}.cabal-install```. This should work in all cases.

jwiegley:
I think @matthewbauer has fixed this in a more future proof way...

PR #325, opened 2018-06-20
Add basic NSelect reduction strategy

picnoir:
Hi,

I tried to create a reduction for NSelect.

Basically, if

- The select is performed on a Set.
- The select attrpath is composed of statickeys.
- This attrpath exists in the set.

The select expression is reduced to the selected expression.

To be honest, I am pretty unsure about this. I still don't get the whole picture.

I have some trouble testing the reduction part of HNix, I am still wondering how this will behave in case of nested sets (ie. `{a={b=1;};}.a.b`.

This is close to be untested: I did not manage find a way to write automated tests for this. How could I do that?

### DIFF
--- a/src/Nix/Reduce.hs
+++ b/src/Nix/Reduce.hs
@@ -53,7 +53,6 @@ import qualified Data.List.NonEmpty as NE
 import           Data.Maybe (fromMaybe, mapMaybe, catMaybes)
 import           Data.Text (Text)
 import           Nix.Atoms
-import           Nix.Convert
 import           Nix.Exec
 import           Nix.Expr
 import           Nix.Frames
@@ -114,7 +113,7 @@ staticImport pann path = do
 
 reduceExpr :: MonadIO m => Maybe FilePath -> NExprLoc -> m NExprLoc
 reduceExpr mpath expr
-  = (`evalStateT` M.empty)
+    = (`evalStateT` M.empty)
     . (`runReaderT` (mpath, emptyScopes))
     . runReducer
     $ cata reduce expr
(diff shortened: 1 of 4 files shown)


picnoir:
In the end I created some unit tests to check the reduction. It seems to behave as I wanted it to behave. 

I am still unsure about the reduction itself, I cannot find any use case for which it may create a problem but I may miss something.

I think this PR is now ready for a review.

src/Nix/Reduce.hs
```diff
@@ -113,7 +113,7 @@ staticImport pann path = do
 
 reduceExpr :: MonadIO m => Maybe FilePath -> NExprLoc -> m NExprLoc
 reduceExpr mpath expr
-    = (`evalStateT` M.empty)
+  = (`evalStateT` M.empty)
```
jwiegley: This is unnecessary...

src/Nix/Reduce.hs
```diff
@@ -155,10 +165,41 @@ reduce (NBinary_ bann op larg rarg) = do
...
+    findBind (x:xs) attrs@(a:|_) = case x of
+        n@(NamedVar (a':|_) _ _) | a' == a -> Just n
+        _                                  -> findBind xs attrs
+    -- Follow the attrpath recursively in sets.
+    inspectSet (NSet_ _ binds) attrs = case findBind binds attrs of
+       Just (NamedVar _ e _) -> case NE.uncons attrs of 
+               (_,Just attrs) -> inspectSet (unFix e) attrs
+               _              -> pure e
+       _ -> sId
+    inspectSet _ _ = sId
```
jwiegley, line 197: All looks good to me, and I like that you've added tests. I'd say this ready for inclusion.

jwiegley (dismissed):

src/Nix/Reduce.hs
picnoir: Indeed, this was unintentional. I'll fix it tomorrow morning.

picnoir: done

Issue #323, opened 2018-06-12
Evaluating 'system' of nixpkgs fails with assertion


infinisil:
Just running

```
hnix --eval -E '(import <nixpkgs> {}).system'
```
gives this error:
```
hnix: /home/infinisil/cfg/external/nixpkgs/lib/systems/parse.nix:252:5:
    |
252 |     assert types.parsedPlatform.check components;
    |     ^
Assertion failed: false
```

- hnix: 0.5.1 from
- nixpkgs: be5e8faafc69195ef47e9cf6b92806cf427656e1

matthewbauer:
I'm debugging this now...

It appears that Hnix doesn't handle equality with functions correctly:

```
let
  f = { a = b: false; };
in

assert f == f;

derivation {
  name = "success";
  builder = "./asd";
  system = "x86_64-linux";
}
```

Works in Nix, but fails in Hnix

Ericson2314:
Hmm I wrote that nixpkgs code, and I don't want it to compare functions like that regardless of the implementation!

jwiegley:
Well, that's interesting. And I can see why, since what `==` sees has already been reduced to a Haskell lambda by that time; we can't "compare expression bodies" anymore by then.

ryantrinkle:
Thankfully, what Nix does seems to not be so bad, here!

```nix
nix-repl> let f = x: x; in f == f 
false
```

From this, and a bit more vague poking around, it appears that functions always compare as unequal.

matthewbauer:
Does it just ignore functions in attr sets then? That would actually be pretty easy to fix in hnix i think.

On the other hand, since it is not very common, it might make sense to just remove this in Nixpkgs and put the assertions somewhere else. I guess that’s kind of cheating though.

matthewbauer:
Yeah so it looks like:

```
$ nix eval '(let f = x: x; in f == f)'
false
$ nix eval '(let a = {f = x: x;}; in a == a)'
true
```

matthewbauer:
Also this:

```
$ nix eval '(let a = [(x: x)]; in a == a)'
true
```

ryantrinkle:
Oh, gross.  Good find, though!

PR #330, opened 2018-07-26
Fix for function equality issues

matthewbauer:
See #323 for more info.


picnoir:
The second commit may generate a conflict with #320.

PR #332, opened 2018-08-01
Fix missing parenthesis in pretty printing selector

### LINKED ISSUE #327: NSelect precedence in pretty printer is broken
```
> prettyNix $ ("x" @@ "y") @. "z"
x y.z
```

That’s because this is commented out:

```haskell
    -- {-  1 -} [ (NSpecialDef "." NSelectOp NAssocLeft,
    --             Postfix $ do
    --                    sel <- seldot *> selector
    --                    mor <- optional (reserved "or" *> term)
    --                    return $ \x -> nSelectLoc x sel mor) ]

    {-  2 -} [ (NBinaryDef " " NApp NAssocLeft,
                -- Thanks to Brent Yorgey for showing me this trick!
                InfixL $ nApp <$ symbol "") ]
```

matthewbauer:
Fixes #327

Another case of missing parenthesis. From the issue:

> prettyNix $ ("x" @@ "y") @. "z"
  x y.z

When it should be:
  (x y).z

Adding non-associative wrapParens seems to fix it. Not sure if this is
the best way but trying a few things & running the test suite still
seem to work.

/cc @jwiegley @Profpatsch

### DIFF
--- a/src/Nix/Pretty.hs
+++ b/src/Nix/Pretty.hs
@@ -202,10 +202,11 @@ exprFNixDoc = \case
     NUnary op r1 ->
         mkNixDoc (text (unpack (operatorName opInfo)) <> wrapParens opInfo r1) opInfo
       where opInfo = getUnaryOperator op
-    NSelect r attr o ->
+    NSelect r' attr o ->
       (if isJust o then leastPrecedence else flip mkNixDoc selectOp) $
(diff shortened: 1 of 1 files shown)


matthewbauer:
Actually this introduces a regression in the tests.

domenkozar:
Could you also add a regression test in `tests/PrettyTests.hs`?

PR #345, opened 2018-08-06
travis: download cachix with smaller closure, pin nixpkgs

domenkozar:
- strict was removed as it's not needed anymore
- cachix in nixpkgs has significantly smaller closure as of last month
- pin nixpkgs to ensure reproducability

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,6 @@ env:
 matrix:
   allow_failures:
     - env: GHCVERSION=ghcjs
-  exclude:
-    - env: GHCVERSION=ghc822 STRICT=false
 
 before_script:
   - sudo mount -o remount,exec,size=4G,mode=755 /run/user || true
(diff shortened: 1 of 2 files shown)


domenkozar:
@jwiegley this is ready :)

jwiegley:
@domenkozar Why is strict no longer needed? We want to make sure that warnings don't creep into the CI build.

domenkozar:
@jwiegley https://github.com/dezgeg/nixpkgs/commit/f3758258b8895508475caf83e92bfb236a27ceb9 was merged to nixpkgs master, which reduces the GCC cli line and thus GHC doesn't fail with `argument list too long`.

If that's the warning you're talking about, this should be fixed.

matthewbauer:
@domenkozar I think there is some confusion here. What you are referring to is ```strictDeps``` which is not implemented in hnix. This STRICT flag just controls whether to evaluate strictly or lazily.

domenkozar:
Oh, I assumed that it's strictDeps yes, I see now that it means -Werror. Reverting.

domenkozar:
@matthewbauer thanks for clarifying my slip, I've pushed changes to exclude that.

jwiegley:
Actually `STRICT` just controls whether we compile with `-Wall` or not; it doesn't affect the semantics of evaluation at all.

PR #349, opened 2018-08-09
Fix function equals test

### LINKED ISSUE #340: Correctly implement function equality testing within structures
This issue tracks the fact that the following tests in `EvalTests.hs` fail when `MATCHING_TESTS=yes`:

  - [ ] `case_function_equals1`
  - [ ] `case_function_equals2`

matthewbauer:
I still don’t understand what Nix is doing behind the scenes here.
This is the original test case that *does* work with my changes.
I've left the broken ones with hopes that hey can be fixed.
Apparently it only happens with let bindings?

Fixes #340


jwiegley (approved):
This looks good to me.

PR #352, opened 2018-09-02
Support tarballs in NIX_PATH and -I

pikajude:
```
$ cabal new-run exe:hnix -- -E 'with import <nixpkgs> {}; pkgs.stdenv.system' --eval --strict -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/9236e9a28d1e084d54ef973ea79cca4156d11347.tar.gz
"x86_64-darwin"
$ NIX_PATH='nixpkgs=https://github.com/NixOS/nixpkgs/archive/9236e9a28d1e084d54ef973ea79cca4156d11347.tar.gz' cabal new-run exe:hnix -- -E 'with import <nixpkgs> {}; pkgs.stdenv.system' --eval --strict
"x86_64-darwin"
```

This only supports tarballs, as does `nix-instantiate`

### DIFF
--- a/src/Nix/Utils.hs
+++ b/src/Nix/Utils.hs
@@ -19,6 +19,7 @@ import qualified Data.HashMap.Lazy as M
 import           Data.List (sortOn)
 import           Data.Monoid (Endo)
 import           Data.Text (Text)
+import qualified Data.Text as Text
 import qualified Data.Vector as V
 import           Lens.Family2 as X
 import           Lens.Family2.Stock (_1, _2)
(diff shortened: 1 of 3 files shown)


src/Nix/Utils.hs
```diff
@@ -108,3 +109,17 @@ toEncodingSorted = \case
                 $ M.toList m
     A.Array l -> A.list toEncodingSorted $ V.toList l
     v -> A.toEncoding v
+
+data NixPathEntryType = PathEntryPath | PathEntryURI deriving (Show, Eq)
```
jwiegley, line 113: I don't think `Nix.Utils` is the place for this type to live... And should we just name is `NixPath`?

jwiegley (changes requested):

PR #360, opened 2018-09-10
Parser exports

infinisil:
- Make `Result` a `Functor`
- Rename some parsers for more consistent exports
- Export useful internal parsers and the `Parser` type

I need these non-full-expression parsers for my nixbot.


src/Nix/Parser.hs
```diff
@@ -14,6 +15,7 @@ module Nix.Parser
     , parseNixText
     , parseNixTextLoc
     , parseFromFileEx
+    , Parser
```
infinisil, line 18: Probably shouldn't do that

PR #363, opened 2018-09-10
More precise freeVars

infinisil:
This should do a much better job at getting the free variables.

Examples:
```nix
freeVars (x: x) # {}
freeVars (x + 1) # {x}
freeVars (with y; x) # {x}
freeVars ({ x ? y.${z} }@z: x) # {y}
freeVars ({ x = 10; y = x; }) # {x}
freeVars (rec { x = 10; y = x; }) # {}
```

Here, a free variable `v` of expression `e` means
- "`v` *may* influence the value of `e`" or
- "`e` *could* depend on `v`"

Reason for the uncertainty is our friend `with`. Because `with y; x` could evaluate just fine if there's an `x` in `y`, and you can override it with `x: with y; x`.

### DIFF
--- a/src/Nix/TH.hs
+++ b/src/Nix/TH.hs
@@ -9,11 +9,12 @@
 module Nix.TH where
 
 import           Data.Fix
-import           Data.Foldable
 import           Data.Generics.Aliases
-import           Data.Set (Set)
+import           Data.Set (Set, (\\))
 import qualified Data.Set as Set
 import qualified Data.Text as Text
+import           Data.List.NonEmpty (NonEmpty(..))
+import           Data.Maybe (mapMaybe)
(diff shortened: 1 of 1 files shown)


jwiegley:
Tests with this would be awesome, but I'll take it. :)

infinisil:
I can provide some more examples like above if you can set up the test infrastructure for them. Never wrote tests myself in Haskell yet, I really should learn that eventually.

infinisil:
Build failure seems unrelated

infinisil:
Oh and I made a mistake in copy pasting something: `freeVars` of `with y; x` are both `y` and `x`, not just `x` as above. The code doesn't make this error, just thought I'd mention it.

PR #369, opened 2018-09-17
Do not remove the context when converting a Set with `outPath` to string

madjar:


shlevy:
@madjar Currently we're not handling context correctly anywhere. See https://github.com/haskell-nix/hnix/pull/289 where we set ourselves up to gradually fix every call site, including these ones.

madjar:
@shlevy Thanks! I figured this was part of a bigger picture, but I went for the quick fix to make my project work.

Shall I close this PR?

shlevy:
I think so, yeah. Curious what use case this fixed though!

madjar:
I have some custom implementation of "nix without a store" that I use for orchestrating batch jobs on HDFS (I'll talk about it at NixCon :) ).

I have my own implementation of `derivationStrict`. The context contains derivations ids (they're hashmap keys instead of store paths), and I had troubles getting the dependencies of a derivation (which I do by grabbing the context of all arguments). This commit fixes that :)

PR #373, opened 2018-11-16
freeVars: Fix inherit's in rec sets and let in

infinisil:
Previously freeVars (rec { inherit y; }) would report nothing, when
really the `y` variable is free. The same for `let inherit y; in y`

Reported by @Synthetica9


Synthetica9 (dismissed):

domenkozar:
I'd really appreciate if such bugs come with a regression test :)

src/Nix/TH.hs
```diff
@@ -69,7 +69,8 @@ freeVars e = case unFix e of
     staticKey (DynamicKey _) = Nothing
 
     bindDefs :: Binding r -> Set VarName
-    bindDefs (Inherit _ keys _) = Set.fromList $ mapMaybe staticKey keys
+    bindDefs (Inherit Nothing _ _) = Set.empty;
```
infinisil, line 72: Technically this aren't the binding definitions anymore, because `inherit x;` does define `x`, but that isn't returned anymore. Maybe it should be named `killSet` instead, from https://en.wikipedia.org/wiki/Reaching_definition

Synthetica9: I don't think it matters too much, seeing it is an internal function that is not exported anywhere. Good names are good though!

Synthetica9:
Oh, my bad that it's currently not building, forgot to add `import qualified Data.Set as S`

tests/EvalTests.hs
```diff
@@ -18,6 +18,7 @@ import           Data.String.Interpolate.IsString
 import           Data.Text (Text)
```
Synthetica9, line 19:
```suggestion
import qualified Data.Set as S
import           Data.Text (Text)
```

Synthetica9 (changes requested):

domenkozar:
There's a merge conflict :)

Synthetica9:
> There's a merge conflict :)

You can just accept both changes, it's two insertions at the bottom of the file

PR #380, opened 2018-11-17
Bump Megaparsec to 7.0.4

vaibhavsagar:


vaibhavsagar:
This doesn't build yet and it looks like all the fun features we used from `Text.Megaparsec.Error` have been changed for the worse now.

vaibhavsagar:
This actually compiles now and all the tests pass, which is strange.

vaibhavsagar:
The error output has changed. This is from `master`:

```bash
$ dist/build/hnix/hnix -E "let a = b in c"
hnix: Parse failed: <string>:1:11:
  |
1 | let a = b in c
  |           ^
unexpected 'i'
expecting ';', ., operator, or the rest of variable binding
```

and this is what it looks like after these changes:

```bash
$ dist/build/hnix/hnix -E "let a = b in c"
hnix: Parse failed: <string>:1:11:
  |
1 | let a = b in c
  |           ^
unexpected 'i'
expecting "!=", "&&", "++", "->", "//", "<=", "==", ">=", "||", '*', '+', '-', '/', ';', '<', '>', '?', or .

```

src/Nix/Render.hs
```diff
@@ -10,26 +10,22 @@
...
 
-renderLocation :: MonadFile m => SrcSpan -> Doc -> m Doc
-renderLocation (SrcSpan beg@(SourcePos "<string>" _ _) _) msg =
-    return $ text $ init $ parseErrorPretty @Char (posAndMsg beg msg)
-
-renderLocation (SrcSpan beg@(SourcePos path _ _) _) msg = do
-    contents <- Nix.Render.readFile path
-    return $ text $ init $ parseErrorPretty' contents (posAndMsg beg msg)
+renderLocation :: Monad m => SrcSpan -> Doc -> m Doc
+renderLocation (SrcSpan beg@(SourcePos _ _ _) _) msg =
```
domenkozar, line 74: I think we still need to handle `<string>` case

domenkozar:
One issue and merge conflicts and I think we're good to go :)

src/Nix/Render.hs
vaibhavsagar: How should it be handled differently?

domenkozar: Ah, seems like path doesn't need resolving anymore and that was the special case. All good :)

vaibhavsagar:
Should be good now.

jwiegley:
@vaibhavsagar Can you resolve the current conflicts and repush? I'd love to merge this in.

vaibhavsagar:
Sure, updated. I'm still building dependencies so I haven't tested this locally just yet but I will repush again if it turns out I broke something when resolving conflicts.

vaibhavsagar:
Confirmed that this successfully builds locally.

vaibhavsagar:
Thanks!

PR #394, opened 2018-11-17
Revert hpack and criterion to fix CI

mightybyte:


mightybyte:
This didn't fix.  In the name of speeding things up this weekend we're going to disable 8.2 in CI for now.

Issue #391, opened 2018-11-17
Make null and empty string not equal


kmicklas:
Compare:

```
hnix> "" == null
true
```

and

```
nix-repl> "" == null
false
```

kmicklas:
Fixes #386.

PR #399, opened 2018-11-17
Remove incorrect cases in valueEq which make empty string and null equal

kmicklas:
This adds on to #391.


shlevy (approved):

mightybyte:
@shlevy Any idea why this is failing one of the language tests?

https://travis-ci.org/haskell-nix/hnix/jobs/456488299#L2254

kmicklas:
The failing test was a bug in the `match` primitive, but now both issues are fixed upstream.

PR #402, opened 2018-11-18
Bump minimum supported version to GHC 8.4.4

jwiegley:
This involved:

  - Changing `default.nix`
  - Changing `.travis.yml`
  - Remove a lot of unnecessary imports, to fix warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ env:
     - secure: "dm6I+M4+V+C7QMTpcSADdKPE633SvmToXZrTbZ7miNDGmMN+/SfHeN2ybi1+PW6oViMlbPN/7J/aEfiGjSJI8vLk72Y4uCWGmpSb8TXZLu6+whnxtZzzW8+z4tsM4048QJg7CF3N/25U8thRFgs3DqUub1Sf3nG9LrNWdz6ZcDQ="
 
   matrix:
-    - GHCVERSION=ghc843 STRICT=false TRACING=false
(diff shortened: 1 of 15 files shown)


mightybyte (approved):
Looks good to me if it passes CI.

jwiegley:
@mightybyte CI broke because `ghc844` requires a newer nixpkgs. I'm going to try pinning to `nixos-18.09`.

vaibhavsagar:
I don't think `nixos-18.09` has `8.4.4`, but I could be wrong.

jwiegley:
@vaibhavsagar https://github.com/NixOS/nixpkgs/blob/release-18.09/pkgs/top-level/haskell-packages.nix seems to indicate that it's there.

domenkozar:
I ported it a couple of days ago (and also 8.6.2)

jwiegley:
@domenkozar OK, so ditch this PR?

jwiegley:
I take it all the warnings have been resolved for 8.6.2?

domenkozar:
The main caveat for GHC 8.6.2 is that it's going to be hard to use nixpkgs haskell set, as it hasn't been updated yet to match stackage nightly. I suggest we stay on GHC 8.4.4 until LTS-13 is released (hopefully in less than a month).

Otherwise my GHC 8.6 branch compiles. I could CPP it so we can supporot 8.4/8.6

jwiegley:
Ok, then I just need to update `cachix` for this to pass CI on Travis.

PR #405, opened 2018-11-19
Remove {To,From}{Nix,Value} Text instances

mightybyte:


src/Nix/Builtins.hs
```diff
@@ -254,15 +254,15 @@ builtinsList = sequence [
...
+    , add' Normal   "stringLength"               (arity1 $ Text.length . principledStringIgnoreContext)
     , add' Normal   "sub"                        (arity2 ((-) @Integer))
     , add' Normal   "substring"                  substring
     , add  Normal   "tail"                       tail_
     , add0 Normal   "true"                       (return $ nvConstant $ NBool True)
     , add  TopLevel "throw"                      throw_
     , add' Normal   "toJSON"
-      (arity1 $ decodeUtf8 . LBS.toStrict . A.encodingToLazyByteString
-                           . toEncodingSorted)
+      (arity1 $ principledMakeNixStringWithoutContext . decodeUtf8 . LBS.toStrict
```
shlevy: The JSON string output of `toJSON` includes the contexts of all of the serialized strings in the value. See https://github.com/NixOS/nix/blob/2f8255cba1597aa1d7afefcac0a33bec63e0ceb6/src/libexpr/value-to-json.cc#L12-L13

src/Nix/Builtins.hs
```diff
@@ -508,29 +508,40 @@ splitDrvName s =
     in (Text.intercalate sep namePieces, Text.intercalate sep versionPieces)
 
 parseDrvName :: forall e m. MonadNix e m => m (NValue m) -> m (NValue m)
-parseDrvName = fromValue >=> \(s :: Text) -> do
+parseDrvName = fromValue >=> fromStringNoContext >=> \(s :: Text) -> do
```
shlevy: Probably don't need this type annotation now

src/Nix/Builtins.hs
```diff
@@ -706,9 +719,11 @@ genericClosure = fromValue @(AttrSet (NThunk m)) >=> \s ->
...
 replaceStrings :: MonadNix e m => m (NValue m) -> m (NValue m) -> m (NValue m) -> m (NValue m)
 replaceStrings tfrom tto ts =
-    fromNix tfrom >>= \(from :: [Text]) ->
-    fromNix tto   >>= \(to   :: [Text]) ->
-    fromValue ts  >>= \(s    :: Text) -> do
+    fromNix tfrom >>= \(nsFrom :: [NixString]) ->
+    fromNix tto   >>= \(nsTo   :: [NixString]) ->
+    fromValue ts  >>= \(ns    :: NixString) -> do
+        let from = map principledStringIgnoreContext nsFrom
+        let to = map principledStringIgnoreContext nsTo
```
shlevy: I think we want to just leave `to` as a list of `NixString`, we want to include the relevant context if we actually do a replacement. See https://github.com/NixOS/nix/blob/2f8255cba1597aa1d7afefcac0a33bec63e0ceb6/src/libexpr/primops.cc#L1988-L1994 and https://github.com/NixOS/nix/blob/2f8255cba1597aa1d7afefcac0a33bec63e0ceb6/src/libexpr/primops.cc#L2016-L2017

src/Nix/Builtins.hs
```diff
@@ -764,10 +780,12 @@ functionArgs fun = fun >>= \case
 
 toFile :: MonadNix e m => m (NValue m) -> m (NValue m) -> m (NValue m)
 toFile name s = do
-    name' <- fromValue name
+    name' <- fromStringNoContext =<< fromValue name
     s' <- fromValue s
-    mres <- toFile_ (Text.unpack name') (Text.unpack s')
-    toNix $ Text.pack $ unStorePath mres
+    mres <- toFile_ (Text.unpack name') (Text.unpack $ principledStringIgnoreContext s')
```
shlevy: We definitely don't want to drop the context from s'! toFile needs to be able to handle strings that have context, as that context will be turned into the runtime references of the resulting file

src/Nix/Builtins.hs
```diff
@@ -764,10 +780,12 @@ functionArgs fun = fun >>= \case
...
+    let t = Text.pack $ unStorePath mres
+        sc = StringContext t $ DerivationOutput t
```
shlevy: This isn't a derivation output, it's a plain path.

src/Nix/Builtins.hs
```diff
@@ -913,9 +939,11 @@ hashString algo s = Prim $ do
         _ -> throwError $ ErrorCall $ "builtins.hashString: "
             ++ "expected \"md5\", \"sha1\", \"sha256\", or \"sha512\", got " ++ show algo
 
+-- TODO Double-check this
 placeHolder :: MonadNix e m => m (NValue m) -> m (NValue m)
-placeHolder = fromValue @Text >=> \_ -> do
-    h <- runPrim (hashString "sha256" "fdasdfas")
+placeHolder = fromValue >=> fromStringNoContext >=> \ns -> do
+    h <- runPrim (hashString (principledMakeNixStringWithoutContext "sha256")
+                             (principledMakeNixStringWithoutContext ns))
```
shlevy: Should be an error if `ns` has context.

src/Nix/Builtins.hs
```diff
@@ -1025,8 +1053,10 @@ addErrorContext _ action = action
 exec_ :: forall e m. MonadNix e m => m (NValue m) -> m (NValue m)
 exec_ xs = do
   ls <- fromValue @[NThunk m] xs
-  xs <- traverse (fromValue @Text . force') ls
-  exec (map Text.unpack xs)
+  xs <- traverse (coerceToString DontCopyToStore CoerceStringy <=< force') ls
+  -- TODO Still need to do something with the context here
+  -- See prim_exec in nix/src/libexpr/primops.cc
+  exec (map (Text.unpack . principledStringIgnoreContext) xs)
```
shlevy: Let's leave this as `hacky` for now?

shlevy (changes requested):

src/Nix/Builtins.hs
mightybyte: Good catch

src/Nix/Builtins.hs
mightybyte: Fixed below

src/Nix/Builtins.hs
mightybyte: The `fromStringNoContext` function does error.  The name `ns` is misleading because it actually has type `Text`.  I'll fix the name.

src/Nix/Builtins.hs
mightybyte: Done

src/Nix/Builtins.hs
```diff
@@ -750,7 +751,8 @@ replaceStrings tfrom tto ts =
                             , Builder.singleton h
                             ]
                     _ -> go rest $ result <> Builder.fromText replacement
-        toNix $ principledModifyNixContents (\x -> go x mempty) ns
+        let ctx = HS.unions $ principledGetContext ns : map principledGetContext nsTo
```
shlevy: Not quite right... We only want to add in the context if the replacement actually happens.

src/Nix/String.hs
```diff
@@ -130,3 +135,7 @@ principledModifyNixContents f (NixString s c) = NixString (f s) c
 -- | Create a NixString using a singleton context
 principledMakeNixStringWithSingletonContext :: Text -> StringContext -> NixString
 principledMakeNixStringWithSingletonContext s c = NixString s (S.singleton c)
+
+-- | Create a NixString using a singleton context
```
shlevy: Copy-paste error

src/Nix/String.hs
```diff
@@ -139,3 +140,12 @@ principledMakeNixStringWithSingletonContext s c = NixString s (S.singleton c)
 -- | Create a NixString using a singleton context
 principledMakeNixString :: Text -> S.HashSet StringContext -> NixString
 principledMakeNixString s c = NixString s c
+
+principledContextToReferences :: NixString -> S.HashSet StringContext
```
shlevy: What's this?

shlevy (changes requested):

src/Nix/String.hs
mightybyte: The start of my implementation of this:

https://github.com/NixOS/nix/blob/master/src/libexpr/primops.cc#L1009

shlevy: Oh, we don't need that, we have explicit `ContextFlavors` instead.

mightybyte:
I've made some progress hunting down the cause of the remaining failing tests.  The `eval-okay-context` test failure goes away when I re-add the following lines (178-181 and 191-194):

https://github.com/haskell-nix/hnix/pull/405/files#diff-3823c33e1d01a3124497afdc267d1f22L178
https://github.com/haskell-nix/hnix/pull/405/files#diff-3823c33e1d01a3124497afdc267d1f22L191

Also, I suspect the `eval-okay-search-path` failure has something to do with these changes to `foldNixPath`:

https://github.com/haskell-nix/hnix/pull/405/files#diff-93c5a26d7401e8c135cdca8f91f36c44R302

Still haven't figured out an appropriate fix yet.

I think I am coming away from all this with the thought that these changes are definitely moving us in a positive direction.  I'm starting to wonder if we should also remove the `{To,From}{Nix,Value} NixString` instances as well.

jwiegley:
@shlevy Is this ready to merge now?

mightybyte:
@jwiegley Not yet.  I still have to address the rest of @shlevy's comments.

src/Nix/Builtins.hs
mightybyte: Fixed

src/Nix/Builtins.hs
mightybyte: Fixed.

src/Nix/String.hs
mightybyte: Removed.

mightybyte:
@shlevy I think this is ready to merge now.  Let me know if there's anything I missed.

shlevy:
@mightybyte Do you want to use the context writer from #422 ?

mightybyte:
@shlevy Sure, but at this point I just want to get this PR merged.

src/Nix/Builtins.hs
```diff
@@ -705,39 +721,44 @@ genericClosure = fromValue @(AttrSet (NThunk m)) >=> \s ->
 
 replaceStrings :: MonadNix e m => m (NValue m) -> m (NValue m) -> m (NValue m) -> m (NValue m)
```
shlevy, line 722: Oof this one is complicated. We should come back to it and document the logic once all of the context stuff is in place.

src/Nix/Builtins.hs
```diff
@@ -763,10 +784,15 @@ functionArgs fun = fun >>= \case
...
     s' <- fromValue s
-    mres <- toFile_ (Text.unpack name') (Text.unpack s')
-    toNix $ Text.pack $ unStorePath mres
+    -- TODO Using hacky here because we still need to turn the context into
+    -- runtime references of the resulting file.
+    -- See prim_toFile in nix/src/libexpr/primops.cc
+    mres <- toFile_ (Text.unpack name') (Text.unpack $ hackyStringIgnoreContext s')
+    let t = Text.pack $ unStorePath mres
+        sc = StringContext t DirectPath
+    toNix $ principledMakeNixStringWithSingletonContext t sc
```
shlevy, line 795: Might be nice to have a function that takes a path and returns the string with singleton context corresponding to that path... later :smile:

src/Nix/Utils.hs
```diff
@@ -123,3 +127,26 @@ uriAwareSplit = go where
                 let ((suffix, _):path) = go (Text.drop 3 e2)
                  in (e1 <> Text.pack "://" <> suffix, PathEntryURI) : path
             | otherwise -> (e1, PathEntryPath) : go (Text.drop 1 e2)
+
+printHash32 :: ByteString -> Text
```
shlevy, line 131: Should eventually dedup this with the logic in hnix-store cc @imalsogreg

shlevy (dismissed):

mightybyte:
@jwiegley This has no conflicts with `pending` but is failing CI.  However, it has conflicts against `master`.  To you want me to keep this PR against `pending` and fix CI or retarget it to master?

jwiegley:
Just needs to pass CI, but looks good to go.

mightybyte:
@jwiegley Fixed.

PR #409, opened 2018-11-20
add typeof and partial binding tracking

domenkozar:
```
hnix> :t 3
an integer
```


Synthetica9:
I'll take a look

jwiegley:
It sound like some of your `m`s elsewhere in the type also need to be `Repl m`.

domenkozar:
Managed to get this to compile and it works:

```
hnix> :t 3
an integer
```

Also fixed GHC 8.6 support (although `ref-tf` dependency needs bounds to be relaxed).

PR #413, opened 2018-11-23
Accumulate string context in builtins.toXML

kmicklas:
I'm not sure if exposing `NixString(..)` is the right thing to do but currently there's no way to get the context out. Encouraging pattern matching over a context-getting function seems like it might actually reduce the risk of forgetting to make use of the context though.

### DIFF
--- a/src/Nix/String.hs
+++ b/src/Nix/String.hs
@@ -2,7 +2,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# OPTIONS_GHC -Wno-unused-top-binds #-}
 module Nix.String (
-    NixString
+    NixString(..)
   , principledMempty
   , StringContext(..)
   , ContextFlavor(..)
--- a/src/Nix/Builtins.hs
+++ b/src/Nix/Builtins.hs
@@ -980,8 +980,7 @@ fromJSON = fromValue >=> \encoded ->
         Right v -> toValue v
 
 toXML_ :: MonadNix e m => m (NValue m) -> m (NValue m)
-toXML_ v = v >>= normalForm >>= \x ->
-    pure $ nvStr $ hackyMakeNixStringWithoutContext $ Text.pack (toXML x)
+toXML_ v = v >>= normalForm >>= pure . nvStr . toXML
 
 typeOf :: MonadNix e m => m (NValue m) -> m (NValue m)
 typeOf v = v >>= toNix @Text . \case
--- a/src/Nix/XML.hs
+++ b/src/Nix/XML.hs
@@ -1,50 +1,68 @@
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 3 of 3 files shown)


src/Nix/String.hs
```diff
@@ -2,7 +2,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# OPTIONS_GHC -Wno-unused-top-binds #-}
 module Nix.String (
-    NixString
+    NixString(..)
```
shlevy, line 5: We were pretty intentionally not exporting this and I'm pretty hesitant to do so here. It should require explicit thought about context to handle strings internally. Is there a way to rewrite this PR without this? Maybe in terms of mappend, or possibly exporting some monad transformer here?

shlevy (commented):
This was already merged, but I'd really like to find another way if possible.

src/Nix/String.hs
domenkozar: Maybe add a comment there that it shouldn't be exported? :)

mightybyte: I already encountered another place where I needed to get the context out and then construct a new NixString with a passed-in context.  That's the reason I created these functions:

https://github.com/haskell-nix/hnix/pull/405/files#diff-3a3776a3f4cdb0350fa1206cf782c14cR58

https://github.com/haskell-nix/hnix/pull/405/files#diff-3a3776a3f4cdb0350fa1206cf782c14cR140

At the end of the day there really aren't very many options.  Either you export the things that are needed by outside code or you add more complex functionality to this module that does all the things needed by outside functions.  That's king of what I was going for with this:

https://github.com/haskell-nix/hnix/pull/405/files#diff-3a3776a3f4cdb0350fa1206cf782c14cR143

shlevy: I guess this is ultimately a @jwiegley question... Do we want to guard against context misusage with specialized names/types or do we want to expect internal users to do the right thing and just export the constructors?

ryantrinkle:
Should we have something like WithStringContext = Writer StringContext a
wherein we're allowed to unpack strings and do computations with them, but
which doesn't allow us to lose the context?  runWithStringContext ::
WithStringContext Text -> NixString

kmicklas:
Yeah I was thinking of something like `extractStringWritingContext :: MonadWriter (HashSet StringContext) m => NixString -> m Text`, but @ryantrinkle's suggestion also makes it impossible to forget the context when running the monad (though it's assuming we always want to produce a `NixString` in the end).

ryantrinkle:
Yep - the assumption that we want a NixString in the end is what lets us
bottle up the effects and prevent accidentally losing things.

shlevy:
Hmm with the writer idea we could even constrain all accesses (except "ignore context" or "force no context" maybe?) to occur within the writer somehow? I mean you're not always creating a new string with context at the end but this might cover most of our cases.

ryantrinkle:
Yes, and we could make primitives that explicitly throw away context.

src/Nix/String.hs
jwiegley: I'd prefer using types to guard against misuse.

PR #418, opened 2018-11-25
Revert "Merge pull request #413 from haskell-nix/xml-string-context"

shlevy:
Let's try to find a way to avoid exporting the NixString constructors
first.

This reverts commit ed073845630a47d53f1b2219336a74ce4993c56c, reversing
changes made to 4bfe2cc60c3add9fcade8e462c5d5357db07bfc4.


shlevy:
See my comments on #413 , might be worth getting on a quick call/chat to hash this out?

jwiegley:
I'll agree with @shlevy and revert this until we've come to a collective decision.

jwiegley:
@shlevy Looks like this broke the build, I'll take a look later tonight.

PR #422, opened 2018-12-06
Accumulate string context in builtins.toXML

kmicklas:
This is a repeat of #413, but this time without exposing the internals of `NixString`.

It's possible we'll find that `WithStringContext` isn't general enough to use in other places eventually, but I think it's a simple solution that works well enough for now.


shlevy (dismissed):
Nice!

jwiegley (changes requested):
Just needs to pass CI.

kmicklas:
@jwiegley It passes now.

PR #427, opened 2018-12-13
Use WithStringContext(T) for nvalueToJSON

kmicklas:


Ericson2314 (dismissed):

src/Nix/Json.hs
```diff
@@ -1,38 +1,45 @@
...
+  . nvalueToJSON
+
+nvalueToJSON :: MonadNix e m => NValue m -> WithStringContextT m A.Value
+nvalueToJSON = \case
+  NVConstant (NInt n) -> pure $ A.toJSON n
+  NVConstant (NFloat n) -> pure $ A.toJSON n
+  NVConstant (NBool b) -> pure $ A.toJSON b
+  NVConstant NNull -> pure $ A.Null
+  NVStr ns  -> do
+    addStringContext $ principledGetContext ns
```
shlevy: Why wouldn't you just do `extractNixString` here?

src/Nix/String.hs
```diff
@@ -145,13 +150,27 @@ principledMakeNixString :: Text -> S.HashSet StringContext -> NixString
 principledMakeNixString s c = NixString s c
 
 -- | A monad for accumulating string context while producing a result string.
-newtype WithStringContext a = WithStringContext (Writer (S.HashSet StringContext) a)
-  deriving (Functor, Applicative, Monad, MonadWriter (S.HashSet StringContext))
+newtype WithStringContextT m a = WithStringContextT (WriterT (S.HashSet StringContext) m a)
```
shlevy, line 153: No need to do this at this stage, but if we're going the mtl route we should consider whether we want a class here.

src/Nix/Json.hs
kmicklas: Woops, oversight.

shlevy (approved):

PR #452, opened 2019-03-09
Nix.Builtins: add getContext and appendContext primops

vaibhavsagar:


jwiegley (changes requested):
Can you please add tests for these two new builtins?

jwiegley (dismissed):

jwiegley (dismissed):

tests/eval-compare/builtins.appendContext.nix
```diff
@@ -0,0 +1,28 @@
+let
```
shlevy, line 1: Why do we need this test when we have the upstream one?

mightybyte: Oh, good point.  @jwiegley requested tests and I guess I  momentarily forgot that we have NixLanguageTests.

jwiegley:
@vaibhavsagar I've merged this by hand, and updated it to match some recent changes to the code. However, I've had to comment out the main bits of the `appendContext` test, since it's not matching the output from `nix-instantiate`. Can you take a look?

vaibhavsagar:
cc @kmicklas

PR #462, opened 2019-03-10
Use fresh name supply to implement pointer equality for thunks

kmicklas:


Ericson2314 (approved):

Ericson2314:
What's that last revert about? Seems harmless.

Issue #442, opened 2019-03-09
Add test to verify string context behavior of builtins.fromJSON


mightybyte:

Anton-Latukha:
Aggregated into #1036.

PR #465, opened 2019-03-10
FromJSON tests

emmanueldenloye:
I created a currently failing test which I think exposes an oversight in the JSON parser in hnix. The failing case is as follows: If there is a key which contains a string that consists of digits, such as "4275", then that key is parsed as a number instead of as a string. This causes the parsed JSON to not match the result that is parsed from nix.

This references issue #442.


jwiegley (changes requested):
Contains unrelated changes...

jwiegley:
Looking good!

jwiegley (approved):

Issue #437, opened 2019-03-09
Add test to verify string context behavior of builtins.pathExists


mightybyte:

hSloan:
I'll pick this one up

PR #470, opened 2019-03-10
WIP: add test for pathExists

hSloan:
Solve: https://github.com/haskell-nix/hnix/issues/437


tests/eval-compare/builtins.pathExists.nix
```diff
@@ -0,0 +1 @@
+builtins.pathExists "/bin/sh/invalid-director"
```
jwiegley: Does this path exist on all machines?

jwiegley (changes requested):

jwiegley:
BTW, GitHub now supports "draft" pull requests, so that you don't need to mark your PR with "WIP" anymore.

jwiegley (changes requested):
If the goal is just a directory that doesn't exist, why about `/foo/bar/baz` or something? I'm not sure whether `/var/empty` needs to exist for this test to work or not... ?

tests/eval-compare/builtins.pathExists.nix
hSloan: No, I saw the previous test assert that a path does not exist.

hSloan:
@jwiegley I've made use of `var/empty` here, this seems to be a convention suggested by @shlevy

PR #485, opened 2019-03-19
Change Eval.hs to use an AlgM instead of an Alg

jwiegley:
Currently there are 123 tests failing, likely due to broken strictness.


jwiegley:
This isn't going to work, since `cataM` begins evaluation from the bottom up, and we need to establish the bindings in a let, for example, before executing the action that processes a name lookup.

Issue #491, opened 2019-03-27
Docker build fails


kaii-zen:
Hello,

I tried to build the docker image according to the README and:

```
$ docker build -t hnix .
Sending build context to Docker daemon  1.403MB
Step 1/7 : FROM lnl7/nix:2.0
 ---> 000789f8ced3
Step 2/7 : WORKDIR /tmp/build
 ---> Using cache
 ---> 40d501c881be
Step 3/7 : COPY default.nix /tmp/build
 ---> Using cache
 ---> 0e718d3b5992
Step 4/7 : COPY package.yaml /tmp/build
COPY failed: stat /var/lib/docker/tmp/docker-builder149832881/package.yaml: no such file or directory
```

vaibhavsagar:
Wow, it looks like that `Dockerfile` is out of date. Could you try my PR at https://github.com/haskell-nix/hnix/pull/492 and see if it fixes this issue for you? I'm building it right now.

vaibhavsagar:
I was mistaken, my PR doesn't yet fix the issue.

Anton-Latukha:
We removed `Dockerfile` like a half or year ago.

The containerization can be rediscussed when HNix can be installed on its own or there would be a pipeline for static compilation, otherwise, we also can resurrect & keep Dockerfile, but it was decided to remove it in the first place, and we would have gargantuan images.

On the merit that there is no Docker support - the Report "Docker build fails" is closed.

kaii-zen:
I honestly don't even remember what this was for so feel free to close.

Anton-Latukha:
@kreisys 
:heart: Thank you for your contribution to the community.

PR #492, opened 2019-03-27
Dockerfile: update

vaibhavsagar:
Refs #491.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,13 +2,11 @@ FROM lnl7/nix:2.0
 
 WORKDIR /tmp/build
 
-COPY default.nix /tmp/build
-COPY package.yaml /tmp/build
+COPY . /tmp/build
 
 # Install tools needed by builtins.fetchTarball, and then install all
 # dependencies into its own layer, which doesn't change.
 RUN nix-env -f '<nixpkgs>' -i gnutar gzip && \
     nix-shell -Q -j2 --run true
 
-COPY . /tmp/build
 RUN bash -xe build.sh
--- a/build.sh
+++ b/build.sh
@@ -4,10 +4,10 @@ set -xe
 set -euo pipefail
 IFS=$'\n\t'
 
-GHCVERSION=${GHCVERSION:-ghc822}
+GHCVERSION=${GHCVERSION:-ghc863}
 STRICT=${STRICT:-false}
 TRACING=${TRACING:-false}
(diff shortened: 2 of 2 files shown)


vaibhavsagar:
Hmm, this doesn't seem to work. I keep getting this error:

```
building '/nix/store/9ll0mifwwr2vxh1zkhqgnarwjgmxg4gn-cabal2nix-hnix.drv'...
installing
error: opening lock file '/nix/var/nix/db/big-lock': Permission denied
error: opening lock file '/nix/var/nix/db/big-lock': Permission denied
** need a revision for VCS when the hash is given. skipping.
** need a revision for VCS when the hash is given. skipping.
** need a revision for VCS when the hash is given. skipping.
** need a revision for VCS when the hash is given. skipping.
cabal2nix: user error (Failed to fetch source. Does this source exist? Source {sourceUrl = "/nix/store/icml732qzxsp88qbsr0b7ab21pilvvhq-build", sourceRevision = "", sourceHash = Guess "", sourceCabalDir = ""})
builder for '/nix/store/9ll0mifwwr2vxh1zkhqgnarwjgmxg4gn-cabal2nix-hnix.drv' failed with exit code 1
```

jwiegley:
Is this PR ready to go yet @vaibhavsagar?

vaibhavsagar:
No, I'm still getting the same error. I think IFD is to blame but I'm not sure. Can you try? It might be something silly that I'm doing.

Ericson2314:
I would just drop Docker, TBH. There shouldn't be hnix-specific about this, vs putting an arbitrary haskell executable in a docker image, right?

Anton-Latukha:
1. `Dockefile`s and their images can be imported in many container specification implementations. Which have are name *plethora*, now.
  For example, `Dockefile` can be imported or directly converted into the Kubernetes configuration.
2. It is a good idea in the container to make a [standalone static binary](https://github.com/nh2/static-haskell-nix), like was [covered on the last NixCon](https://www.youtube.com/watch?v=7hQOiut_MiA).
  Containers&container DevOps'es love static binaries for more lean and trouble-free experience, than a big containers. Which makes `hnix` *AFAIU* a good root to populate some container, if you happen to want/need `hnix` in a container.
3. Also `Dockefile` is good reference documentation on how one sets-up/integrates **this* in the container. Even if the documentation is long, unclear, or not present. Back in time sometimes I preferred to find a good `Dockerfile` to use/look how to set **this* up in a container.

jwiegley:
Dockerfile has been removed.

PR #499, opened 2019-05-16
Make URI's be separate from strings

infinisil:
This makes more sense and allows rewriters to easily see whether a string is an URI or a normal string.

@jwiegley It seems that this was your original intention, as seen from this yet-incorrect comment: https://github.com/haskell-nix/hnix/blob/90f6a61a84fdd41576c0a163d5bec6e55eb2b165/src/Nix/Expr/Types.hs#L104-L106

So I'm not sure if something prevented you from doing this.

All tests still pass, including the ones regarding URI parsing.


infinisil:
Ping?

jwiegley (approved):

jwiegley:
@shlevy What are your thoughts, O master of strings?

Issue #377, opened 2018-11-17
Suggested improvements to NExpr


infinisil:
- [ ] Encoding that `[Binding r]` contains static and dynamic attributes, whereas static ones have to have a different key:
  ```haskell
  data StaticAttributeValue r = Inherit (Maybe r)
                         | NamedVar [NKeyName r] r
  type StaticAttributes r = Map VarName (StaticAttributeValue r)
  type DynamicAttributes r = [(NDynamicAttrPath r, r)]
  type Bindings r = (StaticAttributes r, DynamicAttributes r)
  ```
- [ ] Instead of strings being `[Antiquoted Text r]`, they should be
  ```haskell
  data Interpolated a s = Interpolated a [(s, a)]
  type String = Interpolated Text r
  ```
- [ ] #977
- [x] #1041
- [x] Instead of having `NSet bindings` and `NRecSet bindings`, it should be a single `NSet rec bindings` (first argument indicates it being recursive or not)
- [x] #976

Maybe other improvements are possible too. I really dislike how some things are representable using the current one but never occur in parsing Nix.

Synthetica9:
* Can we also get consistent source positions? Currently some things have their own `SourcePos`, next to the `SrcSpan` provided by `NExprLocF`: https://github.com/haskell-nix/hnix/blob/2f418b26aa92bf79494c899497f327f8c976c814/src/Nix/Expr/Types.hs#L179-L190  Maybe have the annotations be at least parametrised, so we can set them to `()` when it's convenient?

jwiegley:
Should `NAttrPath` take a dynamic boolean argument, rather than two constructors?

jwiegley:
@Infinisil I'm rather open to PRs for these suggestions. Perhaps doing them one at a time would be simplest?

ryantrinkle:
@Synthetica9 Nix has a primitive that retrieves source location for those things *internal to the language*, so they are semantically significant.  The other source locations are used just for error reporting and such.  It may be possible to unify them to some extent, but it's not quite as easy as one would hope.

sjakobi:
https://github.com/haskell-nix/hnix/issues/666#issuecomment-650643370 is somewhat related.

Anton-Latukha:
Just wanted to mention:

> some things are representable ... but never occur in parsing.

Due to Nix paradigms we would never be able to parse everything just by reading the Expr level.

Some things can be made unrepresentable in Expr level, but not all.

[...]

PR #500, opened 2019-05-16
Use NSet for both recursive and non-recursive sets

infinisil:
Simplifies the AST, I suggested this in #377 

To mention is that I didn't change the use of the shorthand functions `mkRecSet` and `mkNonRecSet` for convenience.

I opted to use a new type
```haskell
data NRecordType = NNonRecursive | NRecursive
```

to represent recursive/non-recursiveness. Alternatively a `Bool` could be used, but I like how self-documenting this type is. Constructor names are debatable, `NNonRec/NRec` sounds good too.

This is a backwards-incompatible change.

### DIFF
--- a/src/Nix/Eval.hs
+++ b/src/Nix/Eval.hs
@@ -141,11 +141,11 @@ eval (NList l           ) = do
   scope <- currentScopes
   for l (defer @v @m . withScopes @v scope) >>= toValue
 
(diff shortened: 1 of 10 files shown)


infinisil:
Ping?

jwiegley (approved):

infinisil:
Cool thanks :)

PR #501, opened 2019-05-28
 Effects: Basic: clean-up

Anton-Latukha:
Cleaned from unused language options, while was reading the source.

### DIFF
--- a/src/Nix/Effects/Basic.hs
+++ b/src/Nix/Effects/Basic.hs
@@ -4,7 +4,6 @@
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE MultiWayIf #-}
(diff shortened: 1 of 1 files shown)


jwiegley (approved):

src/Nix/Effects/Basic.hs
```diff
@@ -308,7 +294,7 @@ defaultDerivationStrict = fromValue @(AttrSet (NValue t f m)) >=> \s -> do
     coerceNixList :: NValue t f m -> m (NValue t f m)
     coerceNixList v = do
       xs <- fromValue @[NValue t f m] v
-      ys <- traverse (\x -> demand x coerceNix) xs
```
Ericson2314, line 311: I've never liked this suggestion of hlint. I rather see `flip demand coerceNix`.

Anton-Latukha: It throws me off that you commenting the `master` code. I think you referring to my patch code.
 
I really like
```haskell
(`demand` coerceNix)
```

It is in in code and mentally - one function layer "less". I have good position/block perception so it is easier mentally to feed into the block / drag the block around in the mind, then switch inputs (& that evokes thinking of what types in that inputs are).

It also directly signals that function is used *here* as an infix binary operation, and only one positional argument is fed into it. For me it is easy to process.

Anton-Latukha:
BTW here Hydra Cached failure of some Dependency. Of course I tested compilation the official way inside Nix-shell jail. On NixOS 19.09pre.

If I need something to do - give me a heads-up.

jwiegley (approved):

PR #504, opened 2019-06-15
Replace hashing by cryptonite

sboehler:
ZuriHac project in order to get hnix on stackage

### DIFF
--- a/hnix.cabal
+++ b/hnix.cabal
@@ -495,6 +495,7 @@ library
     , comonad >= 5.0.4 && < 5.1
     , containers >= 0.5.11.0 && < 0.7
     , contravariant >= 1.5 && < 1.6
+    , cryptonite >= 0.25 && <= 0.26
     , data-fix >= 0.2.0 && <  0.3
     , deepseq >=1.4.2 && <1.5
     , dependent-sum >= 0.4 && < 0.5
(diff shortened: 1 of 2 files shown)


flokli:
This helps towards #494.

vaibhavsagar:
This definitely won't work with GHCJS. Is that an issue?

sboehler:
> This definitely won't work with GHCJS. Is that an issue?

I cannot judge on that. But what is the issue with cryptonite and GHCJS?

The alternative approach would be to get hashing itself into stackage.

ryantrinkle:
I'd definitely prefer to keep GHCJS working.

vaibhavsagar:
@sboehler see https://github.com/haskell-crypto/cryptonite/issues/243#issuecomment-421006819.

sboehler:
If hnix is to preserve GHCJS compatibility and get into stackage, some dependencies (like hashing and monadlist) need to be added to stackage.

Closing this for now.

flokli:
I guess GHCJS in general isn't a requirement, but being able to run some version of hnix in the browser, like in [A tour of Nix](https://nixcloud.io/tour/?id=1) is a very useful usecase we shouldn't break easily.

Another possible target to run this on might be webassembly, but I guess it'll also take quite some time until we can get haskell libs with c bindings (like cryptonite) compiled to that platform (see https://github.com/tweag/asterius/issues/148 for some pointers). That left aside, getting crypto done safely in haskell and randomness into webassembly are two other problems.

That being said, getting a pure haskell hasing library into stackage and using that looks like the easiest way forward, at least for now.

PR #506, opened 2019-06-15
Concat list null fixes

lheckemann:
Currently, hnix allows concatenating and comparing lists with null. Nix doesn't, hence I've adjusted the type checker and evaluator to reflect that.


lheckemann:
The hydra failure seems to be unrelated.

jwiegley (approved):

Issue #309, opened 2018-05-09
Replace every use of ErrorCall with a custom Exception value


jwiegley:
Rather than pass strings with rendered components to `ErrorCall`, every possible error or piece of information raised within `hnix` should use a specific constructor, retaining as much information as possible about the context of that information.

Anton-Latukha:
I want to do this task on a long time basis. Until the code is covered.

TL;DL: What is the stabilized module that would be easier/better for me to start with?

#### My current level

I have solid basic Haskell knowledge: [study repo](https://github.com/Anton-Latukha/Haskell-Programming-From-First-Principles) (there are literally all exercises from the chapters, done and commented by myself), and [my theory dictionary](https://github.com/Anton-Latukha/haskell-notes).

I am interested to get deeper in error handling to master it. I know we would choose only particular ways of handling.

I can also do QuickCheck testing. I understand the problematics and limits of property-testing, generators and shrinking. Able to use library classes or write custom generators, properties, and abstract around them. If other testing tools are chosen - I would be eager to learn them also.

I love diving into mathematical theories around Haskell as also philosophy and etymology. I love to understand things I know pretty deep.

You can freely use terminology in directions to my work. That way I would understand directly, and you would also help me.

#### The task

So you request code to have data type constructors for errors.

As I understand it also would go great with smart constructors to weed-out Error cases from the normal flow.

What is the stabilized module that would be easier/better for me to start with?

<sub>P.S. I am a sort of fan of yours, because walking my life I found numerous times I walk with the help of your work, your posts, and with your Elisp code, and with the help of org-habit I changed&changing my life to better.</sub>

Anton-Latukha:
Oh, about Nix.
I once wrote new Nix installer (https://github.com/NixOS/nix/pull/1565), that is despite all my efforts is not merged still.
And I am a temperate contributor and maintainer of a couple of packages in NixPkgs: https://github.com/NixOS/nixpkgs/pull/59293.
And also worked as Nix/NixOS/NixOps/Hydra DevOps.
So for Nix language, and processes/terms I also have some understanding, that would allow me during work better and better understand the `hnix` code base.

Anton-Latukha:
Related: https://github.com/haskell-nix/hnix/pull/760#issue-522869532

Anton-Latukha:
Since PR desynced & closed it - task is no longer in the process - report is open to contribution.

[...]

PR #509, opened 2019-07-24
Replacing every use of ErrorCall with a custom Exception value

Anton-Latukha:
*(Status: 2020-04-18: finished agenda, awaiting reviews & merge)*

Exceptions for HNix: https://github.com/haskell-nix/hnix/issues/309 "_Replace every use of ErrorCall with a custom Exception value_".

Completed exception values coverage.

For this PR:
1) Keeping current PR as close to `master` implementation as possible.
2) Doing only what was asked in the request, placing groundwork by providing data constructors for exceptions.
3) Decided to keep constructors in groups by the function that exception occurs in. I weighted-in the code structure and what is needed and decided to use this middle way approach.
4) For constructors, the Exception type instances are to process data constructors and it's data into a printable message. With `throwError $ EFunctionNameExceptionName a b c`
5) This is a beauty that exception types are not represented in the signatures, so their data constructors can be rearranged & renamed at any moment into any structure, allowing further evolving the handling.
6) Because the exceptions are grouped by function (3) - they are occurred in, naming convention naturally occurred, you can guess the name of the type that gathers together exceptions you are interested in with, prefix `E*` is for *exception* (*synchronous*), `EA*` for *exception asynchronous*. So it guesses aka `EA` + `FunctionName` (abiding the *camelCalse* rules).

### DIFF
--- a/src/Nix/Effects/Basic.hs
+++ b/src/Nix/Effects/Basic.hs
@@ -66,6 +66,16 @@ instance (Show v, Typeable v) => Exception (FileError v)
     =  "File '" <> show name <> "' was not found in the Nix search path "
     <> "(add it using $NIX_PATH or -I)."
 
+data NixPathError a
+  = WrongNixPathFormat a
+  deriving Show
+
+instance (Show v, Typeable v) => Exception (NixPathError v)
+ where
+  displayException (WrongNixPathFormat s)
+    =  "__nixPath must be a list of attr sets with 'path' elements, "
+    <> "but received: '" <> show s <> "'."
+
 defaultMakeAbsolutePath :: MonadNix e t f m => FilePath -> m FilePath
 defaultMakeAbsolutePath origPath = do
   origPathExpanded <- expandHomePath origPath
@@ -165,11 +175,7 @@ findPathBy finder l name = do
     Nothing -> case M.lookup "uri" s of
       Just ut -> defer $ fetchTarball ut
       Nothing ->
-        throwError
-          $  ErrorCall
-          $  "__nixPath must be a list of attr sets"
-          ++ " with 'path' elements, but received: "
-          ++ show s
+        throwError $ ErrorCall $ displayException $ WrongNixPathFormat s
 
 fetchTarball
   :: forall e t f m . MonadNix e t f m => NValue t f m -> m (NValue t f m)
--- a/src/Nix/Effects/Basic.hs
+++ b/src/Nix/Effects/Basic.hs
@@ -105,6 +105,9 @@ defaultMakeAbsolutePath origPath = do
           Nothing -> getCurrentDirectory
           Just v  -> demand v $ \case
             NVPath s -> return $ takeDirectory s
+            -- TODO: 2019-08-12: John Ericson (Ericson2314):
+            -- You don't want to call displayException here, but where the m is
+            -- eliminated.
             v -> throwError $ ErrorCall $ displayException $ CurFileIsntPath v
       pure $ cwd <///> origPathExpanded
   removeDotDotIndirections <$> canonicalizePath absPath
--- a/src/Nix/Effects/Basic.hs
+++ b/src/Nix/Effects/Basic.hs
@@ -62,8 +62,9 @@ instance (Show v, Typeable v) => Exception (EADefaultMakeAbsolutePath v)
     =  "When resolving relative path, __cur_file is in scope, "
     <> "but is not a path; it is: '" <> show v <>"'."
 
-newtype EAFindPathByFile a
+data EAFindPathByFile a
   = EFindPathByFileNotInNixPath a
+  | EFindPathByWrongNixPathFormat a
   deriving Show
 
 instance (Show v, Typeable v) => Exception (EAFindPathByFile v)
@@ -71,14 +72,7 @@ instance (Show v, Typeable v) => Exception (EAFindPathByFile v)
   displayException (EFindPathByFileNotInNixPath name)
     =  "File '" <> show name <> "' was not found in the Nix search path "
     <> "(add it using $NIX_PATH or -I)."
-
-newtype EANixPath a
-  = EWrongNixPathFormat a
-  deriving Show
-
-instance (Show v, Typeable v) => Exception (EANixPath v)
- where
-  displayException (EWrongNixPathFormat s)
+  displayException (EFindPathByWrongNixPathFormat s)
     =  "__nixPath must be a list of attr sets with 'path' elements, "
     <> "but received: '" <> show s <> "'."
 
@@ -214,7 +208,7 @@ findPathBy finder l name = do
     Nothing -> case M.lookup "uri" s of
       Just ut -> defer $ fetchTarball ut
       Nothing ->
-        throwError $ ErrorCall $ displayException $ EWrongNixPathFormat s
+        throwError $ EFindPathByWrongNixPathFormat s
 
 fetchTarball
   :: forall e t f m . MonadNix e t f m => NValue t f m -> m (NValue t f m)
--- a/src/Nix/Effects/Basic.hs
+++ b/src/Nix/Effects/Basic.hs
@@ -52,12 +52,12 @@ import           GHC.DataSize
 #endif
 #endif
 
-data FileError a
+data FileAsyncException a
   = CurFileIsntPath a
   | FileNotInNixPath a
   deriving Show
 
-instance (Show v, Typeable v) => Exception (FileError v)
+instance (Show v, Typeable v) => Exception (FileAsyncException v)
  where
   displayException (CurFileIsntPath v)
     =  "When resolving relative path, __cur_file is in scope, "
@@ -66,23 +66,23 @@ instance (Show v, Typeable v) => Exception (FileError v)
     =  "File '" <> show name <> "' was not found in the Nix search path "
     <> "(add it using $NIX_PATH or -I)."
 
-data NixPathError a
+data NixPathAsyncException a
   = WrongNixPathFormat a
   deriving Show
 
-instance (Show v, Typeable v) => Exception (NixPathError v)
+instance (Show v, Typeable v) => Exception (NixPathAsyncException v)
  where
   displayException (WrongNixPathFormat s)
     =  "__nixPath must be a list of attr sets with 'path' elements, "
     <> "but received: '" <> show s <> "'."
 
-data FetchTarballError a
+data FetchTarballAsyncException a
   = NoUrlAttr
   | NorUriNorSet a
   | NorUriNorString a
   deriving Show
 
-instance (Show v, Typeable v) => Exception (FetchTarballError v)
+instance (Show v, Typeable v) => Exception (FetchTarballAsyncException v)
  where
   displayException NoUrlAttr
     = "builtins.fetchTarball: Missing url attribute."
@@ -199,7 +199,7 @@ fetchTarball
 fetchTarball = flip demand $ \case
   NVSet s _ -> case M.lookup "url" s of
     Nothing ->
-      throwError $ ErrorCall $ displayException (NoUrlAttr :: FetchTarballError String)
+      throwError $ ErrorCall $ displayException (NoUrlAttr :: FetchTarballAsyncException String)
     Just url -> demand url $ go (M.lookup "sha256" s)
   v@NVStr{} -> go Nothing v
   v ->
--- a/src/Nix/Effects/Basic.hs
+++ b/src/Nix/Effects/Basic.hs
@@ -283,10 +283,7 @@ defaultImportPath path = do
         eres <- parseNixFileLoc path
         case eres of
           Failure err ->
-            throwError
-              $ ErrorCall
-              . displayException
-              $ EDefaultImportPathParse
+            throwError $ EDefaultImportPathParse
               $ fillSep ["Parse during import failed:", err]
           Success expr -> do
             modify (M.insert path expr)
--- a/src/Nix/Effects/Basic.hs
+++ b/src/Nix/Effects/Basic.hs
@@ -95,6 +95,7 @@ instance (Show v, Typeable v) => Exception (EAFetchTarball v)
 
 data EADefaultImportPath a
   = EDefaultImportPathParse a
+  | EDefaultImportPath a
   deriving Show
 
 instance (Show a, Typeable a) => Exception (EADefaultImportPath a)
@@ -102,6 +103,8 @@ instance (Show a, Typeable a) => Exception (EADefaultImportPath a)
   -- displayException :: EADefaultImportPath a -> String
   displayException (EDefaultImportPathParse a)
     = show a
+  displayException (EDefaultImportPath a)
+    = "While importing file " <> show a
 
 defaultMakeAbsolutePath :: MonadNix e t f m => FilePath -> m FilePath
 defaultMakeAbsolutePath origPath = do
@@ -272,8 +275,7 @@ defaultImportPath
   -> m (NValue t f m)
 defaultImportPath path = do
   traceM $ "Importing file " ++ path
-  -- TODO: 2019-08-13: Incorporate into exceptions structure
-  withFrame Info (ErrorCall $ "While importing file " ++ show path) $ do
+  withFrame Info (ErrorCall $ displayException $ EDefaultImportPath path) $ do
     imports <- get
     evalExprLoc =<< case M.lookup path imports of
       Just expr -> pure expr
--- a/src/Nix/Effects/Basic.hs
+++ b/src/Nix/Effects/Basic.hs
@@ -93,6 +93,16 @@ instance (Show v, Typeable v) => Exception (EAFetchTarball v)
     = "builtins.fetchTarball: Expected URI or string, received: '"
     <> show v <> "'."
 
+data EADefaultImportPath a
+  = EDefaultImportPathParse a
+  deriving Show
+
+instance (Show a, Typeable a) => Exception (EADefaultImportPath a)
+ where
+  -- displayException :: EADefaultImportPath a -> String
+  displayException (EDefaultImportPathParse a)
+    = show a
+
 defaultMakeAbsolutePath :: MonadNix e t f m => FilePath -> m FilePath
 defaultMakeAbsolutePath origPath = do
   origPathExpanded <- expandHomePath origPath
@@ -272,9 +282,10 @@ defaultImportPath path = do
         case eres of
           Failure err ->
             throwError
-              -- TODO: 2019-08-13: Incorporate into exceptions structure
               $ ErrorCall
-              . show $ fillSep ["Parse during import failed:", err]
+              . displayException
+              $ EDefaultImportPathParse
+              $ fillSep ["Parse during import failed:", err]
           Success expr -> do
             modify (M.insert path expr)
             pure expr
--- a/src/Nix/Effects/Basic.hs
+++ b/src/Nix/Effects/Basic.hs
@@ -52,44 +52,44 @@ import           GHC.DataSize
 #endif
 #endif
 
-data FileAsyncE a
-  = CurFileIsntPath a
-  | FileNotInNixPath a
+data EAFile a
+  = ECurFileIsntPath a
+  | EFileNotInNixPath a
   deriving Show
 
-instance (Show v, Typeable v) => Exception (FileAsyncE v)
+instance (Show v, Typeable v) => Exception (EAFile v)
  where
-  displayException (CurFileIsntPath v)
+  displayException (ECurFileIsntPath v)
     =  "When resolving relative path, __cur_file is in scope, "
     <> "but is not a path; it is: '" <> show v <>"'."
-  displayException (FileNotInNixPath name)
+  displayException (EFileNotInNixPath name)
     =  "File '" <> show name <> "' was not found in the Nix search path "
     <> "(add it using $NIX_PATH or -I)."
 
-data NixPathAsyncE a
-  = WrongNixPathFormat a
+newtype EANixPath a
+  = EWrongNixPathFormat a
   deriving Show
 
-instance (Show v, Typeable v) => Exception (NixPathAsyncE v)
+instance (Show v, Typeable v) => Exception (EANixPath v)
  where
-  displayException (WrongNixPathFormat s)
+  displayException (EWrongNixPathFormat s)
     =  "__nixPath must be a list of attr sets with 'path' elements, "
     <> "but received: '" <> show s <> "'."
 
-data FetchTarballAsyncE a
-  = NoUrlAttr
-  | NorUriNorSet a
-  | NorUriNorString a
+data EAFetchTarball a
+  = ENoUrlAttr
+  | ENorUriNorSet a
+  | ENorUriNorString a
   deriving Show
 
-instance (Show v, Typeable v) => Exception (FetchTarballAsyncE v)
+instance (Show v, Typeable v) => Exception (EAFetchTarball v)
  where
-  displayException NoUrlAttr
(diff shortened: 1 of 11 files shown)


src/Nix/Effects/Basic.hs
```diff
@@ -52,6 +52,14 @@ import           GHC.DataSize
...
 #endif
 
+data EffectsException a
+  = CurFileNotPath a
+  | FileNotInNixPath a
+  | NixPathIsWrong a
+  deriving Show
+
+
+
```
jwiegley, line 64: These new exceptions will need to be rendered, no?

Anton-Latukha:
1) Should I make a small constructors in place that mostly just throw Error and avoid matching the function resulting type. And so not changing the main structure.
2) Or should I propagate Either to the functions, also exported, and as so - make a breaking changes, and align the code in the project with the changes.

Ericson2314:
I don't think this has much use while the text is still thrown at the end of the day. But it you changed it to throw an error type that might be useful.

Rather thinking about this as some dank effects system thing, maybe keep a really simple goal in mind like internationalization. The error type shouldn't need any type parameter, just enough info to write the error messages. Then you can move all the error -> text in one place.

ryantrinkle:
Internationalization would be fantastic.  It's a clear, achievable goal
that could really positively impact users (once the rest of hnix is up to
the task of doing real work)

Anton-Latukha:
Thank you.

---
Because I read somewhere >10 articles, processing some in detail, and seen many approaches. Monadic handling and not, and many arguments.
And having on top to work in existing established project code base got me completely puzzled what to do here.

I really waited for some direction.

---

Throwing error type, abstract `error -> text` to further achieve internationalization - is more details. Witch would help me. And thank you once more.

Ericson2314:
Yeah I figured you had read a bunch of articles. Many such articles tend to advocate very fancy things without first saying why smaller steps are no good. Throwing the error type and rending to strings just for the UI is a good smaller step that's beneficial in itself (the internationalization) and usually compatible with whatever the articles are saying, should we want to do what they suggest later.

Good luck!

Anton-Latukha:
Yes. Also I complain about the situation that Exceptions and Error in standard libs are all mixed-up. It is a bummer that Haskell terminology for own sake does not discern between Exception and Error. Yes - they intertwine, but:

Exception case can be recovered - that means - it (with very high probability) can be recovered though the pure structure.

While unrecoverable Error at the least - throws-out the state, & also most probably better be handled in unpure monad. At least, I most believed the main arguments shown by extensive research in ["Exceptionally Monadic Error Handling" by Jan Malakhovski](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=2&ved=2ahUKEwiPv9nE5PPjAhXlsIsKHaueAJgQFjABegQIAxAC&url=https%3A%2F%2Farxiv.org%2Fpdf%2F1810.13430&usg=AOvVaw0jGLu35gPnM-nXs5Reo22D).

Since Haskell has the pure and unpure realm - the differentiation between Exception and Error seems important to me. Accounting that I am really a rookie.

Anton-Latukha:
About type variable in data type constructor:

`NValue t f m` is polymorphic thunk end-to-end (but it can Show), that means - in the data constructors I can not really point-out the concrete type of it - so that requires polymorphic variable(s) declared in the data type.

Ericson2314:
I don't think you need to show that error type? the error -> error message function will print out enough information.

I would at least try to forge ahead spiking the `Show Error` instance for now, even if you want to return to it later.

src/Nix/Effects/Basic.hs
```diff
@@ -64,13 +105,7 @@ defaultMakeAbsolutePath origPath = do
...
           Just v  -> demand v $ \case
             NVPath s -> return $ takeDirectory s
-            v ->
-              throwError
-                $  ErrorCall
-                $  "when resolving relative path,"
-                ++ " __cur_file is in scope,"
-                ++ " but is not a path; it is: "
-                ++ show v
+            v -> throwError $ ErrorCall $ displayException $ CurFileIsntPath v
```
Ericson2314: You don't want to call `displayException` here, but where the `m` is eliminated.

Anton-Latukha: Do you mean to make a call out of the `do` block/blocks, or when `-> m FilePath` would be reduced? Surely I am missed something understanding your recommendation.

I would mark your point TODO right now, and dodge it to the further cycle over code. And while then maybe you would write or myself would get more understanding on the Exception process.

Since current approach can be drop-in mass-produced in the code I would first do this.

Anton-Latukha: Your suggestions helped me.

Ericson2314: Put most simply, there should be no `ErrorCall` in the code base. If you grep for `ErrorCall` you should get nothing.

I'm sorry I didn't put it this way before. I didn't know `ErrorCall` was a type from elsewhere, and not defined in hnix.

src/Nix/String/Coerce.hs
```diff
@@ -38,6 +38,15 @@ data CopyToStoreMode
   -- ^ Add paths to the store as they are encountered
   deriving (Eq,Ord,Enum,Bounded)
 
+newtype CoerseAsyncException a
```
rycee: `CoerseAsyncException` → `CoerceAsyncException`

Anton-Latukha: Thank you, `rycee`, you already helped me several times in the community.

src/Nix/Effects/Basic.hs
Anton-Latukha: Yes.

```haskell
-- | This is thrown when the user calls 'error'. The first @String@ is the
-- argument given to 'error', second @String@ is the location.
data ErrorCall = ErrorCallWithLocation String String
```
currently it seems to be the default constructor for such purposes (currently [`base GHC.Exception`](https://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.Exception.html#line-66) considered the main path).

Ericson2314: Keep in mind throwError comes from the MonadError class and is used purely here (I think). There's no Haskell exceptions from this stuff.

Anton-Latukha: Ok. The Haskell vocabulary and situation is completely perplexing the communication here.
Probably you are saying that the process is pure, until the computation arrives outside into IO monad. Of course - strictly by particularly Haskell theory that it is true, unpurity contained in the IO monad.

`Throw` in combination with `error` terms themselves by the name state that it must end not pure in some cases (Error/Exception or Async Exception/Sync Exception) of terminology I know of.
If throws are not caught - they go into IO monad in the end and trigger the effects. And the majority of situations, of these exception constructors, are not recoverable (unpure), and some are possible to recover with request of IO input (unpure), and the messages are there to be printed/logged, which you guessed...

Yes, the code inside IO monad is effectful and moves into C, for what I know. But if this pure functions are most directly exist to raise the effects - I think if we want we can debate between philosophy, definitions an applications of pure/unpure terms indefinitely in this functions.

Anton-Latukha: But I just basically restated that this is that everyone knows in this thread.

From Haskell terms of pure realm - yes, `throwM` & `MonadThrow` are pure by themselves.
But again - these Exception situations are not recoverable, so their evaluation results into IO effects.

Thanks for raising my attention to this, indeed I was not mentally distinguishing the IO Rubicon line.

Ericson2314: https://hackage.haskell.org/package/mtl-2.2.2/docs/Control-Monad-Except.html#v:throwError see the instances for that class. there is the IOException instance which uses exceptions but I don't think we are using that.

Anton-Latukha: The `throwError` is redefined in HNix:
https://github.com/haskell-nix/hnix/blob/c03a40aceac5221c82603c9208b381a76c72100f/src/Nix/Frames.hs#L53-L58

GHCI:
```haskell
*Nix> :l src/Nix.hs
...
*Nix> :i throwError
throwError ::
  (Framed e m, Exception s, Control.Monad.Catch.MonadThrow m) =>
  s -> m a
  	-- Defined at src/Nix/Frames.hs:55:1
```
`toException` [is](https://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.Exception.Type.html#toException):
```haskell
class (Typeable e, Show e) => Exception e where
    toException   :: e -> SomeException
    fromException :: SomeException -> Maybe e

    toException = SomeException
    fromException (SomeException e) = cast e
```

& when `throwM` is launched - it shortens the monadic computation.

There is a bunch of custom HNix type classes & instances extending the IO monad in `src/Nix/Effects.hs`:
https://github.com/haskell-nix/hnix/blob/c03a40aceac5221c82603c9208b381a76c72100f/src/Nix/Effects.hs#L86-L113

`Effects.hs` has a second largest Exception counter over all HNix files, I am moving towards it. Also it is the root of all evil, so I decided to learn the simplier, more isolated, abstracted modules before.

I just already learned a bunch of project code while working on it. Also normal tooling fitting to the project helps here: https://github.com/Anton-Latukha/haskell-notes#integration-of-nixosnix-with-haskell-ide-engine-hie-and-emacs-spacemacs.

Anton-Latukha: Also I see (as also you can see here) there is too much forward and backward transformations of Exceptions, these computations seems to be can be reduced.

Ericson2314: My apologies I didn't realize we were doing those things. @jwiegley has the final say, but I would ditch Exception and SomeException and only use closed error types. I really don't like open error types.

Anton-Latukha: No apologies, as I do errors everyone does them.

Currently I'm just doing a bunch of Exception constructors as was asked.
I am gathering a list of major point questions to ask/tweak during the merge preparation.

Ericson2314: In that case you will want to implement the `Exception` class so you can `throwError $ YourNewNiceType arg0 ...argn` directly without any `ErrorCall`.

Anton-Latukha:
I think to name the exceptions in form of types `EWhereWhat`, data `EWhereWhat`, and document what that means. `EA` acronym would be for asynchronous exceptions. 
Since HNix already uses a lot of types with acronym prefixes: `NTypes`, `TTypes`, like `NVConstant` `NVBuiltin`...

Anton-Latukha:
The good news - I covered most of the source cases with needed Exception constructors, currently 58/92=63% of ErrorCalls.

All which are left (34) are `Builtins.hs` file.

Anton-Latukha:
@jwiegley 

Here all ErrorCall exceptions and all other exception occurrences received a unique value.

This PR is ready to take the review.

Anton-Latukha:
@haskell-nix

Someone, please review / merge.

This is pretty simple PR in itself, a bunch of constructors.

I want to lay this groundwork and evolve it further, with the account of your suggestions.

jwiegley (approved):

src/Nix/Effects/Basic.hs
jwiegley: @Ericson2314 is right, instead of `ErrorCall $ displayException $ EMonadExecMissingProgram u u`, it should just be `EMonadExecMissingProgram u u`. The rendering into a string part should only happen at reporting time.

jwiegley (changes requested):
See feedback in the currently open comment.

src/Nix/Effects/Basic.hs
Anton-Latukha: Please, I need a bit more direct information to understand properly.

At this stage structures are drop-in replacements, they are equivalent to strings. In this PR the agenda is to lift the Exception handling from the hardcode into data type constructions, which was requested, allowing use of type class & function handling, which opens ability for the normal Haskell development.

Currently the delta is aimed at to be the first step & to be drop-in straight - using `displayException` produce the `String` that is equivalent to the `String`s earlier hardcoded and do only that. That is why currently data types are so minimal & have only `Show` type class and only it, AFAIK they not to require anything other. We ask the collected info to be `Show, Typeable` and then with it compose `String`, and that is why this constructors need only `Show`, to be an equivalent `String` as currently this was an agenda.

PR is already too large.

After merge I would do a new up to date literature review on the topic, there were valuable articles published, as well as projects created. Then I would open a brief discussion. Then would, as was requested, gradually migrate from `ErrorCall`'s to the found solution.

jwiegley: Thanks, @Anton-Latukha, except that I don't think it's worthwhile to create an intermediate state of development in `master` if the intention to migrate toward real exceptions in the near future. That way, other people working on the project won't get confused as to whether they should use `ErrorCall` or a specific exception.

We have time here, so why don't we just add those commits into this PR branch until it takes on the intended final form.

Anton-Latukha: Ok.

Sorry for the silence for a couple of days.

Ok.

Anton-Latukha:
@jwiegley 

I've done a literature review on the topic of Exception handling.

#### Small report

Particularly fresh and good sources are particularly in currently unreleased books ["Haskell In Depth" (Vitaly Bragilevsky)](https://livebook.manning.com/book/haskell-in-depth/chapter-7/) and ["Intermediate Haskell" (chapter on the topic by Mark Karpov)](https://markkarpov.com/tutorial/exceptions.html).

I've reviewed a number of other sources and sorted-out the terminology and started to have ideas about theory.

Overall the solutions are agreeably moving towards `Control.Exception.Safe`.

#### Agenda

I currently try to understand why HNix Exception stack has:

https://github.com/haskell-nix/hnix/blob/9f1519d9ca7d37b6584b8eaabde0ca47c65857cd/src/Nix/Frames.hs#L53-L58

in the form that it is.

For my task, this need to be reconstructed/modified by me to integrate nicely with a more general approach of handling Exceptions.

The code everywhere currently does:

```haskell
throwError $ ErrorCall $ displayException EConstructor
```

And looking into the `throwError` internally and it's external usage we see the back-n-forth transformation of types that needs to be reduced.

And you ask to develop handling further to the proper form, which directly involves `throwError`.

One of the last uncovered questions in my knowledge for solid base of modern Haskell - are lenses. It is the last question I still need to cover.

`throwError` is a method in `Frames.hs` and directly relies on lenses. And that I've could sort-out if I was not puzzled by the `Frames.hs` in the first place, I still try to understand the concept of what are these "frames" fully are. As I understand it holds the "frame of reference" and representation, it seems to gather computation information and regard output to Syslog standard (but not quite, since there are no "Notify" level introduced, name points to be internal Nix-specific signaling levels).

I'd wanted to have some explanation or documentation for Frames first.

To be able to develop I see a need a space to be able to maneuver, to rebase `throwError` on `Control.Exception.Safe`. As I currently see the `throwError` as currently created is for the throwing of errors that should be not recovered (since uses `Error :: NixLevel`), `Warning` level messages/methods are not present, `Debug` and `Info` are sometimes logged as:
```haskell
withFrame Debug (ErrorCall $ displayException EMapAttrs)
```

1. I want to ask some phrases/documentation on Frames.
2. Some phrases on maneuvering on `{throwError, withFrame, logging}` on refactoring to build-in the `{bracket*, throw*, catch*/handle*}` system into the project. Probably the good idea is to have an internal handling system in the `{bracket*, throw*, catch*/handle*}` form, gather needed info with it and then catch and rethrow into HNix/Frame modified form, in that way code would be a lot simpler and understandable, we would use the intuitive and known `throw` and use `throwError` once in the code to output in the needed form.

Anton-Latukha:
And at current point, I mostly needed to just reduce the `throwError $ ErrorCall $ displayException EConstructor` -> `throwError EConstructor`. That is the landslide of cases, so I would request merging this.

There are some `ErrorCall`s left, but they are due to specific uses in more complex blocks of code, or blocks that better be discussed separately.

Anton-Latukha:
It took me some deeper reading and understanding to destinct roughly *synchronous* form *asynchronous* exceptions. In the end, the most solid, clear and right terminology description is in the initial source that introduced *synchronous* / *asynchronous* exception terms into Haskell terminology - ["Asynchronous Exceptions in Haskell" (Simon Marlow, Simon Peyton Jones, Andrew Moran, John Reppy)](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/07/asynch-exns.pdf).

By the right terminology, all exceptions in the project which I touched (`ErrorCall's` and throws - are all *synchronous* exceptions), so I PR code in accordance.

src/Nix/Effects/Basic.hs
Anton-Latukha: Yep, reduced that.

jwiegley:
I think this work is really close to completion. Maybe sometime if we could pair program it together, it would only take an hour or two to hammer out. Please feel free to reach out to `johnw` on IRC (#haskell) or jwiegley on gitter.im (hnix has its own room there).

Anton-Latukha:
@jwiegley Thank you. I am gladly taking this proposition.

I am living and growing longterm, the same longterm is a wish to serve and develop in this project.

Also as I actively asked and received maintenance rights. Now, when I took it - I should keep and carry thoroughly some of the maintenance roles, to be with the world fully.

So on the contact, you can also at once give me all instructions and advice you would decide to give.

jwiegley:
@Anton-Latukha For my own part I just want to get myself out of the way of your willingness and enthusiasm to aid the project, so consider me available to support and encourage however you need. If you want to know more about where things stand now, or what future ideas might be most in line with the project's goals, please ping me on IRC or gitter.im, or e-mail me, and I'll make myself available.

sjakobi:
I would like to see this work cross the finishing line, but I'm not sure how I can help. My experience with and my understanding of Haskell exceptions is admittedly limited.

Personally, I'm surprised to see so many different error _types_ here. In the code bases that I'm familiar with, there usually are rather few different error types but each may have many constructors. See for example [`dhall`'s `TypeMessage`](http://hackage.haskell.org/package/dhall-1.33.1/docs/Dhall-TypeCheck.html#t:TypeMessage) or the [`dhall-json` `CompileError`](http://hackage.haskell.org/package/dhall-json-1.6.4/docs/Dhall-JSON.html#t:CompileError).

My suspicion is that the multitude of different error types here might make it difficult to catch the exceptions `hnix` produces and might also make the types of some functions much more complex.

But that's just my impression – it would be better to get an assessment from someone who has more experience with this project.

Anton-Latukha:
Well, I several times asked for help on showing me the way to do it properly. And I kind of already knew back then - there is no proper way. And now I understand that the silence shows that the others also do not know what to do about it.

During the work read a bunch about the error handling. And I wrote a bunch summarizing it also.
Error handling only started to standardize around now, practices still not settled.

Overall there is no decision or recommendation to keep error data types unique, grouped or to hold everything in one monstrous error data type.

I've chosen the middle-way grouping I comed-up with that is easy to start with. Then we able to add info to them. Then the main thing in this way - we can progress either way, grouping them is of course possible.

Over time and on the promotion, and since I did not got any particular suggestions on continuation, I of course lost some interest in this PR, since there is other important stuff to attend to, and this is somewhat supplementary to the project. And on top of it all since it is big PR it is a quite heavy question addressing it and get permission to merge it. Because it big - it of course aslo desyncs quickly. The old story. I love to merge code and then to work on it more, you know, I belabored it many times already, otherwise, even the creator of the PR works but feels there is no progress in the question.

There is basically only two-three places that are not covered in PR, because this PR is nicely structured, and in those 2-3 which left - there is additional changes needed on top of the code.

Anton-Latukha:
+====

I would commit to rebase this in near future.

What I want is:
1. Boilerplate should be accepted structurally as is. So it gets in and available for further work on it.
It would already be marginally better than throwing ErrorCalls in place.:
  What was requested - is to `"
Replace every use of ErrorCall with a custom Exception value"` #309 - and this is what PR does already, not counting for 2 calls that I plan to cover at the final stage, or in interaction, after PR merged
  What I want to understand here, is after the merge of boilerplate - what structure to put on top and how to abstract it better further.
  
2. I for the N-th time ask from more experienced people what structure they think fit to put over this boilerplate, so at least we can put in place the TODO Report, and we would approach the HNix as effectively as we can with all resources we have.

sjakobi:
@Anton-Latukha 

> I would commit to rebase this in near future.

I'd suggest that you coordinate the way forward for this PR with @jwiegley directly before investing more time into this. Maybe try reaching him via email if he doesn't respond here.

To me personally, it's unfortunately not obvious that this PR is an improvement in the direction that #309 so vaguely pointed at. So I don't feel comfortable with merging this without approval from @jwiegley or someone else who is familiar with the original plan.

In general, I would advise you not to create huge PRs without having some confirmation that you're taking the right approach.

Anton-Latukha:
Well, you can see in above discussion that actually I put the argument that I made research and that direction is not known, asked more experienced to give more clear direction number of times, was one that wanted bite-size PRs, and that I wanted a gradual process, and I tried to force merges, but so far there was no merge, so I kept adding constructors.

So all things you suggested - in fact, were actively pursued here by me. Often the contributor is not at the cause, but he fights with discussion the chosen paradigms of processes that created side-effects.

I just felt uncomfortable to sit down right now and contributing other code, knowing that that code I write in Git would contradict with this unmerged code that became the big PR.

I am Ok with any solution. I am personally totally Ok with not having error handling currently, and see how things would pan-out both in HNix and Haskell Computer Science development.

But the thing is, if someone would do this anew - he would need to do all this boilerplate anyway again, to lift it further.

jwiegley:
@Anton-Latukha Let's see if I can't spend some time with this to show you the form I'm thinking of. It should be a pretty mechanical transformation of what you've already done, so the bulk of what you contributed is good stuff we'll want to keep.

Anton-Latukha:
Frames subsystem got fixed.

Vaginally remember encountering that type problem which was fixed.

The Frame system probably needs to be used for exceptions.

But are all exceptions fit the Frame idea/design/purpose?

Some particular exceptions probably/may not fit the use of Frame.

Anton-Latukha:
During my reading on the exception handling in Haskell, the most thought-provoking article was/is this simple one: https://www.haskellforall.com/2019/12/prefer-to-use-fail-for-io-exceptions.html

People try to classify, divide & conquer exception handling in many ways. But I think this article drawn the ephemeral for current views & practices but and in the future the most important distinction, the exceptions & their types **must** be managed as an ephemeral system that is call on need inside compiler code evaluation system, until there is really no way but not to throw that "one" exception, and that it is precisely when "that exception" should/would be transferred from ephemeral system to the IO (throw/catch or miss).

Currently thinking that all exception handling must be approached and established from this point of view first, before looking in any throw/catch and async stuff. Because how the exception gets thrown is not as much important, if it does not happen to be thrown and so does not need to be stored in the first place, and as soon as it is understood that exception does not apply - its lazy thunk needs to be gone from memory.

It feels like keeping the exceptions out of the explicit type system signatures and give them to GHC which should have solid API type/handling system for them - is the key for the ecosystem to have a good exception story, exceptions need to stay ephemeral until thrown, and when thrown - go under the system and materialize there to be catched. And when they get really thrown & catched - then the type system may again be valuable to manage exceptions information. Seems like the pure code logic and exception management code logic - are two very different systems that need a GHC wormhole between them, to the degree that exception handling needs to be possible to move to the internal module, and currently we often try to drag them both around everywhere and thread through everything by Either. *idk*, still to need to know more.

---

Ok, this time can say that from the patch the `master` diverged so much, it is easier to close it, and redo all ~109 constructors some other time. *idk* does HNix needs all those custom exception constructors, especially before a proper message reporting system and handling of its contents, thinking that the messaging system with exception system inside need to be groomed (or at least understood and documented), before spawning exception constructors for the system that still a bit undefined, needs a bit of formation and coverage across the system and picking the proper data types that get used for it.

PR #512, opened 2019-08-04
Fixing the errors found in messages

Anton-Latukha:


src/Nix/Effects/Basic.hs
```diff
@@ -165,7 +165,7 @@ findPathBy finder l name = do
         throwError
           $  ErrorCall
           $  "__nixPath must be a list of attr sets"
-          ++ " with 'path' elements, but saw: "
+          ++ " with 'path' elements, but recieved: "
```
jwiegley:
```suggestion
          ++ " with 'path' elements, but received: "
```

jwiegley (dismissed):

src/Nix/Effects/Basic.hs
Anton-Latukha: Yep, facepalm. My own spelling still not that great.

jwiegley (approved):

PR #515, opened 2019-08-31
Allow most recent versions of dependencies.

quasicomputational:
The only actual breaking change affecting `hnix` is `these` splitting
into three packages and renaming some things. Otherwise, as the
package should work fine with either the newer or older versions, I
have left the previous bounds in place and added the new ones
disjunctively.

This will help somewhat with #494.

--

A release with these bounds bumps would also be handy, since GHC 8.8 support probably isn't getting backported to old versions of libraries. `hnix` won't work out of the box with 8.8 yet, but only due to dependencies, and it's quite likely there'll be nothing else to do once they catch up.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -8,15 +8,14 @@
 
 , withHoogle  ? true
 
-, rev    ? "ed1b59a98e7bd61dd7eac266569c294fb6b16300"
-, sha256 ? "0b2wdbbaqdqccl7q9gskhfjk7xaqvjwcls4b6218anyc247gscnb"
+, rev    ? "c4adeddb5f8e945517068968d06ea838b7c24bd3"
 
 , pkgs   ?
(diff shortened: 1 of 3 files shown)


quasicomputational:
It might make sense now to re-bound everything, with the lower bound coming from the version in the package set the new Nixpkgs is using. I was hoping to avoid having to bump the versions used in CI, but it makes more sense than a web of overrides and manually solving a build plan.

quasicomputational:
The Linux build timed out; not entirely unexpected, since it's doing a lot of rebuilding. I think it might finish if it's restarted and allowed to pull from the cache.

No clue why the OSX builder is sick, except maybe I grabbed a bad Nixpkgs commit to pin to?

jwiegley (approved):

PR #525, opened 2019-11-24
Fix compound attributes in let bindings

layus:
Fix an issue where compound attributes would discard previous bindings
in a let statement.

`let a=1; b.b=2; c=3; in a` would fail, where
`let a=1; b.b=2; c=3; in c` would work as expected.

The fix is trivial, but we could have a discussion about using `M.insertWith (flip const) k pos p` instead of `M.insert k pos p` to keep the first position where the subset was encountered instead of the last one. There is probably no reason whatsoever to access such a sub-attrset, so the discussion is probably not worth it.

Example:

```
let
  x.a = 1;
  x.b = 2;
in ...
```

Should we consider that x was declared in line 2 or three ? To me, it would make even more sense to say that it was declared nowhere, and use noPos.

### DIFF
--- a/src/Nix/Eval.hs
+++ b/src/Nix/Eval.hs
@@ -207,8 +207,8 @@ attrSetAlter (k : ks) pos m p val = case M.lookup k m of
     ( M.insert
       k
       (toValue @(AttrSet v, AttrSet SourcePos) =<< (, mempty) <$> sequence st')
-      st
-    , M.insert k pos sp
+      m
+    , M.insert k pos p
     )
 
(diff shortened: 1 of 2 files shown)


jwiegley:
Hm.. What is the answer when we have this:

```
let y = { x.a = 1;
  x.b = 2;
}; in y.x
```

Where is `y.x` said to be defined?

layus:
With the above patch, it will be reported as defined 
```
let y = { x.a = 1;
  x.b = 2;
  ^-- here
}; in y.x
```

PR #526, opened 2019-11-25
Add proper default 'nix' prefix to search paths

### LINKED ISSUE #419: Ship nix/fetchurl.nix with hnix
This gets us closer to simple usage :)

layus:
The search path of Nix contains by default the "nix" prefix that points
to $datadir/nix/corepkgs, where $datadir defaults to $prefix/lib at
installation time, but can be overriden by NIX_DATA_DIR.

We implemented it using `Paths.hnix.getDataDir` and `NIX_DATA_DIR` to
follow Nix behaviour as closely as possible.

A small discrepancy is that we do the lookup on each invocation, where
Nix caches the searchPath at context creation.

/edit Oh, and it fixes #419.

### DIFF
--- a/src/Nix/Builtins.hs
+++ b/src/Nix/Builtins.hs
@@ -97,7 +97,7 @@ import           Nix.Value
 import           Nix.Value.Equal
 import           Nix.Value.Monad
 import           Nix.XML
-import           System.Nix.Base32             as Base32
+import           System.Nix.Base32              as Base32
 import           System.FilePath
 import           System.Posix.Files             ( isRegularFile
                                                 , isDirectory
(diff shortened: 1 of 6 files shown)


src/Nix/Builtins.hs
```diff
@@ -97,7 +97,7 @@ import           Nix.Value
 import           Nix.Value.Equal
 import           Nix.Value.Monad
 import           Nix.XML
-import           System.Nix.Base32              as Base32
+import           System.Nix.Base32             as Base32
```
jwiegley: This whitespace change is not needed.

jwiegley (changes requested):

src/Nix/Builtins.hs
layus: Updated.

jwiegley (approved):

layus:
Thanks !

PR #528, opened 2019-11-27
Refactor execBinaryOp to demand values where needed.

### LINKED ISSUE #527: Crash with binary operator: how to force thunks ?
HNix fails on inputs like `1 + builtins.head [3]` with an erro message like this:

```
hnix: In <string>:1:1:
While evaluating:
>>>>>>>>
  Fix
    (NBinary
       NPlus
       (Fix (NConstant (NInt 1)))
       (Fix
          (NBinary
             NApp
             (Fix
                (NSelect (Fix (NSym "builtins")) (StaticKey "head" :| []) Nothing))
             (Fix (NList [ Fix (NConstant (NInt 3)) ])))))
<<<<<<<<
Unsupported argument types for binary operator NPlus: Free (NVConstantF (NInt 1)), Pure <thunk>
```

Evaluating `builtins.head [3]` works correctly with `--force` and shows `<CYCLE>` when evaluated without `--force`. This relates to the problem, because from the source and the error message above, we see that the binary operator "+" cannot handle the thunk generated by `builtins.head`. I spent some time investigating, but this way beyond my current skills.

I am very interested in an explanation of the yet-to-come solution :-)

layus:
This splits execBinaryOp in two parts where the second conatains all
the logic that relies on the assumption that values are already "demanded".

Fixes #527.

\cc @jwiegley

### DIFF
--- a/src/Nix/Exec.hs
+++ b/src/Nix/Exec.hs
@@ -349,134 +349,121 @@ execBinaryOp
   -> m (NValue t f m)
   -> m (NValue t f m)
 
-execBinaryOp scope span NOr larg rarg = fromValue larg >>= \l -> if l
-  then orOp Nothing True
-  else rarg >>= \rval -> fromValue @Bool rval >>= orOp (Just rval)
+execBinaryOp scope span op lval rarg = case op of
+  NEq   -> rarg >>= \rval -> valueEqM lval rval >>= boolOp rval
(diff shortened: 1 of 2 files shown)


layus:
The new design splits by operator, which makes it a bit more clear. For example, I added a case that was not covered before (Paths + String) because it was obviously missing in the list of operator "+".

I also fixed two uses of hackyStrXXX.

jwiegley:
Thank you!

PR #543, opened 2020-04-18
README: code blocks: upd cabal keys to v2-*; unification of markup

Anton-Latukha:
#### 1. `v2-*` should be used now.

Information: https://www.haskell.org/cabal/users-guide/nix-local-build-overview.html

`v1-*,v2-*` - are future-proof compatible keys;
`new-*` is used for new features in the *beta* stage;
```
cabal --help

 [new-style projects (beta)]
  new-build ...

 [new-style projects (forwards-compatible aliases)]
  v2-build ...
```

So from `Cabal 3.0` all Nix `new-*` are `v2-*` and should be them.

#### 2. Upgraded all `nix-shell -> cabal` to use `v2-*`.

Since Cabal runs inside the Nix-shell - it is obvious that `v2-*` features must be used for Cabal to be proper.

(`v2-*` has some breaking changes in keys from `new-*` (such cases examples shown [here](https://github.com/Anton-Latukha/haskell-notes#95-ghcid)), but in case of HNix instructions - it is a drop-in replacement)

I've tested all README instructions with `v2-*` - commands are proper, have proper additional keys and work.

#### 3. All code blocks have one style.

---

The rendered results can be viewed on my [PR branch](https://github.com/Anton-Latukha/hnix/tree/README-2020-04-18-upd).

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ for interacting with store paths, until `hnix-store` is ready.
 
 ## Getting Started
 
-```bash
+```
 $ git clone --recursive https://github.com/haskell-nix/hnix.git
 ...
 $ cd hnix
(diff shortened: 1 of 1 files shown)


jwiegley:
Do people really have to type `v2-` in front of cabal commands these days? It can't just be `cabal configure` and `cabal build`?

PR #546, opened 2020-04-25
remove leftover Poly README/LICENSE

sorki:
Can create a better version with credits / license moved to appropriate places if needed.


Ericson2314:
Yeah maybe just delete the README for now, until we figure out whether any extent code traces itself to that and thus needs attribution.

PR #549, opened 2020-04-27
pretty derivation builtin nix

sorki:
Just a style / readability fix. My mind is now twisted by Haskell 2D syntax..


jwiegley:
Thanks for this and all the other changes!

PR #550, opened 2020-04-27
drop GEq instances

sorki:
Breaks the build on GHC 8.8.3.

Not sure where they belong and what they are good for so if someone can shed some light on this :)


Ericson2314:
What is the error for these? They now go in https://github.com/phadej/some .

sorki:
It was just due to missing `Data.Typeable` import, closing this in favor of https://github.com/phadej/some/pull/34

PR #552, opened 2020-05-08
fix refl dependency

jhartzell42:
dependent-sum no longer re-exports Refl, and this fixes that.


jhartzell42:
Replacing with one from OSOS repo

Issue #472, opened 2019-03-10
Replace hnix's `StorePath` with hnix-store's `Path`


mightybyte:
The hnix-store work is probably still too much in flux (as of March 2019), but this will probably be something we want to do eventually.

sorki:
Assigning my self to this and the rest of the `hacky*` issues.

Anton-Latukha:
`hacky*` are mainly relate around `NixString` type, `StorePath` may be also.

I've opened a largely unanswered question-statement. Proper is to populate `NAtom` with all atomic Nix types that are atomic in the Nix design ("Can NixString be the NAtom?" #534).

Also probably related to the `hacky*` implementations is the "convert nix file to haskell json object" #539.

sorki:
Interesting, I've tried to eval `release-combined.nix` and it looks pretty good except that `drv` paths are different compared to Nix - mostly caused by missing `inputSrcs` and `inputDrvs` from the derivation so I would like to tackle that now and reach parity with Nix.

layus:
A discussion about this was raised in #554: https://github.com/haskell-nix/hnix/pull/554/files#r538011509 && https://github.com/haskell-nix/hnix/pull/554/files#r427298390

PR #554, opened 2020-05-13
derivationStrict

### LINKED ISSUE #364: Implement derivationStrict without nix-instantiate
Once we have string context and hnix-store integration, we can implement derivationStrict properly

layus:
Implement derivationStrict primOp

Closes #364

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -111,19 +111,19 @@
 
 let
 
- getDefaultGHC = "ghc${
-    (
-      # Remove '.' from the string 8.8.4 -> 884
-      pkgs.lib.stringAsChars (c: if c == "." then "" else c)
-        # Get default GHC version,
-        (pkgs.lib.getVersion pkgs.haskellPackages.ghc)
-    )
-  }";
+  getDefaultGHC = "ghc${
+      (
+        # Remove '.' from the string 8.8.4 -> 884
+        pkgs.lib.stringAsChars (c: if c == "." then "" else c)
+          # Get default GHC version,
+          (pkgs.lib.getVersion pkgs.haskellPackages.ghc)
+      )
+    }";
 
- compilerPackage =
-   if ((compiler == "") || (compiler == "default"))
-     then getDefaultGHC
-     else compiler;
+  compilerPackage =
+    if ((compiler == "") || (compiler == "default"))
+      then getDefaultGHC
+      else compiler;
 
   #  2020-12-31: NOTE: Remove after `hnix-store 0.4` arrives into Nixpkgs
   hnix-store-src = pkgs.fetchFromGitHub {
@@ -226,13 +226,14 @@ let
       # 2020-12-07 We really want cryptohash-sha512, but it conflicts with
       # recent versions of base, for seemingly no valid reason.
       # As the update is slow to happen, just jailbreak here
-      # See https://github.com/haskell-hvr/cryptohash-sha512/pull/3
-      # and https://github.com/haskell-hvr/cryptohash-sha512/pull/5
+      # See https://github.com/haskell-hvr/cryptohash-sha512 PRs 3, 5 and issue 4
+      # See also https://github.com/NixOS/nixpkgs/pull/106333 for a temporary fix.
       cryptohash-sha512 = pkgs.haskell.lib.unmarkBroken ( pkgs.haskell.lib.doJailbreak super.cryptohash-sha512 );
 
       # 2020-12-07 hnix-store-remote fails when trying to connect to a real hnix daemon.
       # probably due to nix sandbox restrictions.
-      hnix-store-remote = pkgs.haskell.lib.dontCheck super.hnix-store-remote;
+      # Upstream issue @ https://github.com/haskell-nix/hnix-store/issues/80
+      hnix-store-remote = pkgs.haskell.lib.removeConfigureFlag super.hnix-store-remote "-fio-testsuite";
 
       # 2020-08-04 hnix uses custom LayoutOptions and therefore is
       # likely to be affected by the change in the ribbon width
--- a/default.nix
+++ b/default.nix
@@ -91,7 +91,7 @@
 #   , nixos-20.03  # Last stable release, gets almost no updates to recipes, gets only required backports
 #   ...
 #   }
-, rev ? "24eb3f87fc610f18de7076aee7c5a84ac5591e3e"
+, rev ? "8ba15f6383c74e981d8038fa19cc77ed0c53ba22"
 
 , pkgs ?
     if builtins.compareVersions builtins.nixVersion "2.0" < 0
@@ -125,16 +125,16 @@ let
      then getDefaultGHC
      else compiler;
 
-  #  2020-05-23: NOTE: Currently HNix-store needs no overlay
-  # hnix-store-src = pkgs.fetchFromGitHub {
-  #   owner = "haskell-nix";
-  #   repo = "hnix-store";
-  #   rev = "0.2.0.0";
-  #   sha256 = "1qf5rn43d46vgqqgmwqdkjh78rfg6bcp4kypq3z7mx46sdpzvb78";
-  # };
+  #  2020-12-31: NOTE: Remove after `hnix-store 0.4` arrives into Nixpkgs
+  hnix-store-src = pkgs.fetchFromGitHub {
+    owner = "haskell-nix";
+    repo = "hnix-store";
+    rev = "fd09d29b8bef4904058f033d693e7d928a4a92dc";
+    sha256 = "0fxig1ckzknm5g19jzg7rrcpz7ssn4iiv9bs9hff9gfy3ciq4zrs";
+  };
 
   overlay = pkgs.lib.foldr pkgs.lib.composeExtensions (_: _: {}) [
-    # (import "${hnix-store-src}/overlay.nix")
+    (import "${hnix-store-src}/overlay.nix" pkgs pkgs.haskell.lib)
     (self: super:
       pkgs.lib.optionalAttrs withHoogle {
       ghc = super.ghc // { withPackages = super.ghc.withHoogle; };
@@ -223,6 +223,17 @@ let
     root = packageRoot;
 
     overrides = self: super: {
+      # 2020-12-07 We really want cryptohash-sha512, but it conflicts with
+      # recent versions of base, for seemingly no valid reason.
+      # As the update is slow to happen, just jailbreak here
+      # See https://github.com/haskell-hvr/cryptohash-sha512/pull/3
+      # and https://github.com/haskell-hvr/cryptohash-sha512/pull/5
+      cryptohash-sha512 = pkgs.haskell.lib.unmarkBroken ( pkgs.haskell.lib.doJailbreak super.cryptohash-sha512 );
+
+      # 2020-12-07 hnix-store-remote fails when trying to connect to a real hnix daemon.
+      # probably due to nix sandbox restrictions.
+      hnix-store-remote = pkgs.haskell.lib.dontCheck super.hnix-store-remote;
+
       # 2020-08-04 hnix uses custom LayoutOptions and therefore is
       # likely to be affected by the change in the ribbon width
       # calculation in prettyprinter-1.7.0.
--- a/hnix.cabal
+++ b/hnix.cabal
@@ -402,8 +402,8 @@ library
     , gitrev >= 1.1.0 && < 1.4
     , hashable >= 1.2.5 && < 1.4
     , hashing >= 0.1.0 && < 0.2
-    , hnix-store-core >= 0.3.0 && < 0.4
-    , hnix-store-remote >= 0.2.0 && < 0.3
+    , hnix-store-core >= 0.4.0 && < 0.5
+    , hnix-store-remote >= 0.4.0 && < 0.5
     , http-client >= 0.5.14 && < 0.6 || >= 0.6.4 && < 0.7
     , http-client-tls >= 0.3.5 && < 0.4
     , http-types >= 0.12.2 && < 0.13
--- a/hnix.cabal
+++ b/hnix.cabal
@@ -341,6 +341,7 @@ library
       Nix.Convert
       Nix.Effects
       Nix.Effects.Basic
+      Nix.Effects.Derivation
       Nix.Eval
       Nix.Exec
       Nix.Expr
@@ -401,8 +402,9 @@ library
     , gitrev >= 1.1.0 && < 1.4
     , hashable >= 1.2.5 && < 1.4
     , hashing >= 0.1.0 && < 0.2
-    , hnix-store-core >= 0.1.0 && < 0.3
-    , http-client >= 0.5.14 && < 0.6 || >= 0.6.4 && < 0.8
+    , hnix-store-core >= 0.3.0 && < 0.4
+    , hnix-store-remote >= 0.2.0 && < 0.3
+    , http-client >= 0.5.14 && < 0.6 || >= 0.6.4 && < 0.7
     , http-client-tls >= 0.3.5 && < 0.4
     , http-types >= 0.12.2 && < 0.13
     , lens-family >= 1.2.2 && < 2.2
--- a/hnix.cabal
+++ b/hnix.cabal
@@ -11,9 +11,19 @@ license:        BSD3
 license-file:   LICENSE
 build-type:     Simple
 cabal-version:  >= 1.10
+data-dir:       data/
+data-files:
+    nix/corepkgs/buildenv.nix
+    nix/corepkgs/unpack-channel.nix
+    nix/corepkgs/derivation.nix
+    nix/corepkgs/fetchurl.nix
+    nix/corepkgs/imported-drv-to-derivation.nix
 extra-source-files:
     data/nix/corepkgs/buildenv.nix
+    data/nix/corepkgs/unpack-channel.nix
     data/nix/corepkgs/derivation.nix
+    data/nix/corepkgs/fetchurl.nix
+    data/nix/corepkgs/imported-drv-to-derivation.nix
     data/nix/tests/lang/binary-data
     data/nix/tests/lang/data
     data/nix/tests/lang/dir1/a.nix
--- a/src/Nix/Effects.hs
+++ b/src/Nix/Effects.hs
@@ -7,6 +7,8 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE TypeApplications #-}
 
 module Nix.Effects where
 
@@ -17,23 +19,30 @@ import           Prelude                 hiding ( putStr
 import qualified Prelude
 
 import           Control.Monad.Trans
+import qualified Data.HashSet                  as HS
 import           Data.Text                      ( Text )
 import qualified Data.Text                     as T
-import           Network.HTTP.Client     hiding ( path )
+import qualified Data.Text.Encoding            as T
+import           Network.HTTP.Client     hiding ( path, Proxy )
 import           Network.HTTP.Client.TLS
 import           Network.HTTP.Types
 import           Nix.Expr
-import           Nix.Frames
+import           Nix.Frames              hiding ( Proxy )
 import           Nix.Parser
 import           Nix.Render
 import           Nix.Utils
 import           Nix.Value
 import qualified Paths_hnix
-import qualified System.Directory              as S
 import           System.Environment
 import           System.Exit
+import           System.FilePath                ( takeFileName )
 import qualified System.Info
 import           System.Process
+
+import qualified System.Nix.Hash               as Store
+import qualified System.Nix.Store.Remote       as Store
+import qualified System.Nix.Store.Remote.Types as Store
+import qualified System.Nix.StorePath          as Store
 
 -- | A path into the nix store
 newtype StorePath = StorePath { unStorePath :: FilePath }
@@ -226,36 +235,54 @@ print = putStrLn . show
 instance MonadPutStr IO where
   putStr = Prelude.putStr
 
-class Monad m => MonadStore m where
-    -- | Import a path into the nix store, and return the resulting path
-    addPath' :: FilePath -> m (Either ErrorCall StorePath)
 
-    -- | Add a file with the given name and contents to the nix store
-    toFile_' :: FilePath -> String -> m (Either ErrorCall StorePath)
+type RecursiveFlag = Bool
+type RepairFlag = Bool
+type StorePathName = Text
+type FilePathFilter m = FilePath -> m Bool
+type StorePathSet = HS.HashSet StorePath
+
+class Monad m => MonadStore m where
+
+    -- | Add a path to the store, with bells and whistles
+    addToStore :: StorePathName -> FilePath -> RecursiveFlag -> RepairFlag -> m (Either ErrorCall StorePath)
+    default addToStore :: (MonadTrans t, MonadStore m', m ~ t m') => StorePathName -> FilePath -> RecursiveFlag -> RepairFlag -> m (Either ErrorCall StorePath)
+    addToStore a b c d = lift $ addToStore a b c d
+
+    -- | Add a nar (action) to the store
+    -- addToStore' :: StorePathName -> IO Nar -> RecursiveFlag -> RepairFlag -> m (Either ErrorCall StorePath)
+
+    addTextToStore' :: StorePathName -> Text -> Store.StorePathSet -> RepairFlag -> m (Either ErrorCall StorePath)
+    default addTextToStore' :: (MonadTrans t, MonadStore m', m ~ t m') => StorePathName -> Text -> Store.StorePathSet -> RepairFlag -> m (Either ErrorCall StorePath)
+    addTextToStore' a b c d = lift $ addTextToStore' a b c d
+
+parseStoreResult :: Monad m => String -> (Either String a, [Store.Logger]) -> m (Either ErrorCall a)
+parseStoreResult name res = case res of
+  (Left msg, logs) -> return $ Left $ ErrorCall $ "Failed to execute '" ++ name ++ "': " ++ msg ++ "\n" ++ show logs
+  (Right result, _) -> return $ Right result
 
 instance MonadStore IO where
-  addPath' path = do
-    (exitCode, out, _) <- readProcessWithExitCode "nix-store" ["--add", path] ""
-    case exitCode of
-      ExitSuccess -> do
-        let dropTrailingLinefeed p = take (length p - 1) p
-        pure $ Right $ StorePath $ dropTrailingLinefeed out
-      _ ->
-        pure
-          $  Left
-          $  ErrorCall
-          $  "addPath: failed: nix-store --add "
-          ++ show path
 
-  --TODO: Use a temp directory so we don't overwrite anything important
-  toFile_' filepath content = do
-    writeFile filepath content
-    storepath <- addPath' filepath
-    S.removeFile filepath
-    pure storepath
+  addToStore name path recursive repair = case Store.makeStorePathName name of
+    Left err -> return $ Left $ ErrorCall $ "String '" ++ show name ++ "' is not a valid path name: " ++ err
+    Right pathName -> do
+      -- TODO: redesign the filter parameter
+      res <- Store.runStore $ Store.addToStore @'Store.SHA256 pathName path recursive (const False) repair
+      parseStoreResult "addToStore" res >>= \case
+        Left err -> return $ Left err
+        Right storePath -> return $ Right $ StorePath $ T.unpack $ T.decodeUtf8 $ Store.storePathToRawFilePath storePath
+
+  addTextToStore' name text references repair = do
+    res <- Store.runStore $ Store.addTextToStore name text references repair
+    parseStoreResult "addTextToStore" res >>= \case
+      Left err -> return $ Left err
+      Right path -> return $ Right $ StorePath $ T.unpack $ T.decodeUtf8 $ Store.storePathToRawFilePath path
+
+addTextToStore :: (Framed e m, MonadStore m) => StorePathName -> Text -> Store.StorePathSet -> RepairFlag -> m StorePath
+addTextToStore a b c d = either throwError return =<< addTextToStore' a b c d
 
 addPath :: (Framed e m, MonadStore m) => FilePath -> m StorePath
-addPath p = either throwError pure =<< addPath' p
+addPath p = either throwError return =<< addToStore (T.pack $ takeFileName p) p True False
 
 toFile_ :: (Framed e m, MonadStore m) => FilePath -> String -> m StorePath
-toFile_ p contents = either throwError pure =<< toFile_' p contents
+toFile_ p contents = addTextToStore (T.pack p) (T.pack contents) HS.empty False
--- a/src/Nix/Effects.hs
+++ b/src/Nix/Effects.hs
@@ -244,14 +244,15 @@ type StorePathSet = HS.HashSet StorePath
 
 class Monad m => MonadStore m where
 
-    -- | Add a path to the store, with bells and whistles
+    -- | Copy the contents of a local path to the store.  The resulting store
+    -- path is returned.  Note: This does not support yet support the expected
+    -- `filter` function that allows excluding some files.
     addToStore :: StorePathName -> FilePath -> RecursiveFlag -> RepairFlag -> m (Either ErrorCall StorePath)
     default addToStore :: (MonadTrans t, MonadStore m', m ~ t m') => StorePathName -> FilePath -> RecursiveFlag -> RepairFlag -> m (Either ErrorCall StorePath)
     addToStore a b c d = lift $ addToStore a b c d
 
-    -- | Add a nar (action) to the store
-    -- addToStore' :: StorePathName -> IO Nar -> RecursiveFlag -> RepairFlag -> m (Either ErrorCall StorePath)
-
+    -- | Like addToStore, but the contents written to the output path is a
+    -- regular file containing the given string.
     addTextToStore' :: StorePathName -> Text -> Store.StorePathSet -> RepairFlag -> m (Either ErrorCall StorePath)
     default addTextToStore' :: (MonadTrans t, MonadStore m', m ~ t m') => StorePathName -> Text -> Store.StorePathSet -> RepairFlag -> m (Either ErrorCall StorePath)
     addTextToStore' a b c d = lift $ addTextToStore' a b c d
--- a/src/Nix/Effects/Basic.hs
+++ b/src/Nix/Effects/Basic.hs
@@ -8,10 +8,6 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE ViewPatterns #-}
 
-{-# OPTIONS_GHC -Wno-missing-signatures #-}
-{-# OPTIONS_GHC -Wno-orphans #-}
-{-# OPTIONS_GHC -fno-warn-name-shadowing #-}
-
 module Nix.Effects.Basic where
 
 import           Control.Monad
@@ -20,30 +16,24 @@ import           Data.HashMap.Lazy              ( HashMap )
 import qualified Data.HashMap.Lazy             as M
 import           Data.List
 import           Data.List.Split
-import           Data.Maybe                     ( maybeToList )
 import           Data.Text                      ( Text )
 import qualified Data.Text                     as Text
-import           Nix.Atoms
+import           Data.Text.Prettyprint.Doc
 import           Nix.Convert
 import           Nix.Effects
 import           Nix.Exec                       ( MonadNix
-                                                , callFunc
                                                 , evalExprLoc
                                                 , nixInstantiateExpr
                                                 )
 import           Nix.Expr
 import           Nix.Frames
-import           Nix.Normal
 import           Nix.Parser
-import           Nix.Pretty
 import           Nix.Render
 import           Nix.Scope
 import           Nix.String
-import           Nix.String.Coerce
 import           Nix.Utils
 import           Nix.Value
 import           Nix.Value.Monad
-import           Prettyprinter
 import           System.FilePath
 
 #ifdef MIN_VERSION_ghc_datasize
@@ -126,8 +116,8 @@ findPathBy
   -> [NValue t f m]
   -> FilePath
   -> m FilePath
-findPathBy finder l name = do
-  mpath <- foldM go Nothing l
+findPathBy finder ls name = do
+  mpath <- foldM go Nothing ls
   case mpath of
     Nothing ->
       throwError
@@ -263,39 +253,6 @@ pathToDefaultNixFile :: MonadFile m => FilePath -> m FilePath
 pathToDefaultNixFile p = do
   isDir <- doesDirectoryExist p
   pure $ if isDir then p </> "default.nix" else p
-
-defaultDerivationStrict
-  :: forall e t f m . MonadNix e t f m => NValue t f m -> m (NValue t f m)
-defaultDerivationStrict = fromValue @(AttrSet (NValue t f m)) >=> \s -> do
-  nn <- maybe (pure False) (demand ?? fromValue) (M.lookup "__ignoreNulls" s)
-  s' <- M.fromList <$> mapMaybeM (handleEntry nn) (M.toList s)
-  v' <- normalForm =<< toValue @(AttrSet (NValue t f m)) @_ @(NValue t f m) s'
-  nixInstantiateExpr $ "derivationStrict " ++ show (prettyNValue v')
- where
-  mapMaybeM :: (a -> m (Maybe b)) -> [a] -> m [b]
-  mapMaybeM op = foldr f (pure [])
-    where f x xs = op x >>= (<$> xs) . (++) . maybeToList
-
-  handleEntry :: Bool -> (Text, NValue t f m) -> m (Maybe (Text, NValue t f m))
-  handleEntry ignoreNulls (k, v) = fmap (k, ) <$> case k of
-      -- The `args' attribute is special: it supplies the command-line
-      -- arguments to the builder.
-      -- TODO This use of coerceToString is probably not right and may
-      -- not have the right arguments.
-    "args"          -> demand v $ fmap Just . coerceNixList
-    "__ignoreNulls" -> pure Nothing
-    _               -> demand v $ \case
-      NVConstant NNull | ignoreNulls -> pure Nothing
-      v'                             -> Just <$> coerceNix v'
-   where
-    coerceNix :: NValue t f m -> m (NValue t f m)
-    coerceNix = toValue <=< coerceToString callFunc CopyToStore CoerceAny
-
-    coerceNixList :: NValue t f m -> m (NValue t f m)
-    coerceNixList v = do
-      xs <- fromValue @[NValue t f m] v
-      ys <- traverse (`demand` coerceNix) xs
-      toValue @[NValue t f m] ys
 
 defaultTraceEffect :: MonadPutStr m => String -> m ()
 defaultTraceEffect = Nix.Effects.putStrLn
--- a/src/Nix/Effects/Basic.hs
+++ b/src/Nix/Effects/Basic.hs
@@ -225,13 +225,13 @@ findPathM = findPathBy existingPath
     pure $ if exists then Just apath else Nothing
 
 defaultImportPath
-  :: (MonadNix e t f m, MonadState (HashMap FilePath NExprLoc) m)
+  :: (MonadNix e t f m, MonadState (HashMap FilePath NExprLoc, b) m)
   => FilePath
   -> m (NValue t f m)
 defaultImportPath path = do
   traceM $ "Importing file " ++ path
   withFrame Info (ErrorCall $ "While importing file " ++ show path) $ do
-    imports <- get
+    imports <- gets fst
     evalExprLoc =<< case M.lookup path imports of
       Just expr -> pure expr
       Nothing   -> do
@@ -242,7 +242,7 @@ defaultImportPath path = do
               $ ErrorCall
               . show $ fillSep ["Parse during import failed:", err]
           Success expr -> do
-            modify (M.insert path expr)
+            modify (\(a, b) -> (M.insert path expr a, b))
             pure expr
 
 defaultPathToDefaultNix :: MonadNix e t f m => FilePath -> m FilePath
--- a/src/Nix/Effects/Derivation.hs
+++ b/src/Nix/Effects/Derivation.hs
@@ -16,9 +16,11 @@ import           Prelude                 hiding ( readFile )
 import           Control.Arrow                  ( first, second )
 import           Control.Monad                  ( (>=>), forM, when )
 import           Control.Monad.Writer           ( join, lift )
+import           Control.Monad.State            ( MonadState, gets, modify )
 
 import           Data.Char                      ( isAscii, isAlphaNum )
 import qualified Data.HashMap.Lazy             as M
+import qualified Data.HashMap.Strict           as MS
 import qualified Data.HashSet                  as S
 import           Data.List
 import qualified Data.Map.Strict               as Map
@@ -100,7 +102,7 @@ writeDerivation (drv@Derivation {inputs, name}) = do
 
 -- | Traverse the graph of inputDrvs to replace fixed output derivations with their fixed output hash.
 -- this avoids propagating changes to their .drv when the output hash stays the same.
-hashDerivationModulo :: (Framed e m, MonadFile m) => Derivation -> m (Store.Digest 'Store.SHA256)
+hashDerivationModulo :: (MonadNix e t f m, MonadState (b, MS.HashMap Text Text) m) => Derivation -> m (Store.Digest 'Store.SHA256)
 hashDerivationModulo (Derivation {
     mFixed = Just (Store.SomeDigest (digest :: Store.Digest hashType)),
     outputs,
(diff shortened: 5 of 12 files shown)


jwiegley:
This looks like great work, @kmicklas, I wonder if @shlevy and @ryantrinkle might have some time to take a look. If not, we can merge later next week and fix whatever is left to be fixed.

kmicklas:
I think you meant to tag @layus :smile:

jwiegley:
Oh, right, thank you @kmicklas !

src/Nix/Effects.hs
```diff
@@ -226,36 +242,58 @@ print = putStrLn . show
...
+        Left error -> return $ Left error
+        Right path -> return $ Right $ StorePath $ T.unpack $ T.decodeUtf8 $ Store.storePathToRawFilePath path
+
+  addTextToStore' name text references repair = do
+    -- TODO: Handle references
+    res <- Store.runStore $ Store.addTextToStore name text references repair
+    path <- parseStoreResult "addTextToStore" res
+    case path of
+      Left error -> return $ Left error
+      Right path -> return $ Right $ StorePath $ T.unpack $ T.decodeUtf8 $ Store.storePathToRawFilePath path
```
sorki, line 280: `storepath` branch now has `storePathToText` iff needed but it would be better to switch to `StorePath` type of `hnix-store-core`.

src/Nix/Effects.hs
```diff
@@ -226,36 +242,58 @@ print = putStrLn . show
...
-    writeFile filepath content
-    storepath <- addPath' filepath
-    S.removeFile filepath
-    return storepath
+  addToStore name path recursive repair = case Store.makeStorePathName name of
+    Left error -> return $ Left $ ErrorCall $ "String '" ++ show name ++ "' is not a valid path name: " ++ error
+    Right pathName -> do
+      -- todo: pass the hash explicitly
+      -- TODO: Handle the filter
+      res <- Store.runStore $ Store.addToStore pathName path recursive (Proxy :: Proxy Store.SHA256) (const False) repair
```
sorki: This also changed type to omit proxy so you can just do `Store.addToStore @'Store.SHA256 pathName ..`

sorki:
https://github.com/haskell-nix/hnix-store/pull/59 now also contains parser for `Derivation`s but needs https://github.com/Gabriel439/Haskell-Nix-Derivation-Library/pull/8

There's also support for building derivations now if you want to give it a try.

layus:
Build failure due to the dependencies on newer versions of the hnix-store... We should realy integrate everything in one single repo.

sorki:
Now buildable from `hnix-overlay` with GHC 8.10 using merged branches - https://github.com/sorki/hnix-overlay/commit/4f3454b53198348e35289b0fca6985ab2fe4df98

Seems to work but eats all my ram :smile: 

Be careful as this does access real system store via daemon now (tests do as well).

layus:
This PR is very inefficient. We need a way to cache hashDerivationModulo results. Otherwise we are doomed with an exponential computing time. It means that in practice computing `nixpkgs.hello.drvPath` does not complete in reasonable time on my machine.

It may also be related to the memory consumption issue a we load several times all the derivations from memory, parse everything and compute hashes. enforcing some strictness may also help.

From my point of view, this state should reside in the store itself. We should considr moving some of the code to hnix-store.

Anton-Latukha:
@layus Above message may decerve to be an idea, so it is good to preseve it in Report.

layus:
Hi, I think this has reached a mergeable state. :tada: :tada: :tada: 

The main issue with exponential memory usage (in the size of the closure) is now addressed by adding a cache for hashDerivationModulo results, next to the cache of parsed nix files (see last commit). The bulk of the implementation resides in the second to last commit. Lots of code, but standalone for the most part.
The rest of the commits are refactoring, fixes and changes in various locations. Some of these changes are not strictly needed for this PR. If they are deemed undesirable, I can of course redact them.

There are no tests, but nixpkgs.hello gives the same hash with nix and hnix. Achieving this was quite a feat :tada: and requires the fixes to hnix-store that are idling in https://github.com/haskell-nix/hnix-store/pull/64. Also, tracking down the issue with the substring builtin was quite a pain.

Anyway, I think this is quite an achievement as it bridges the gap between hnix an nix. We can now correctly evaluate derivations ! :1st_place_medal: 

/cc @jwiegley, @shlevy, @ryantrinkle Would you mind having a look at this PR ?

layus:
Pending questions to be resolved later on, in other PRs:

- [ ] Consider moving hashDerivationModulo to hnix-store. It would also require to move the cache there.
- [ ] Decide which store we want by default in hnix (readonly, standalone, daemon).
- [ ] Implement proper back-and-forth communication between MonadStore (hnix-store) and MonadNix++ (hnix). @imalsogreg made a great job of tossing monads around for streaming nars in https://github.com/haskell-nix/hnix-store/pull/61. Maybe he could give some advice here too.
    - [ ] Implement a proper path filter based on a nix function.
    - [ ] Keep a single daemon socket for the whole hnix lifetime. Currently we open a new one on each store operation.
    - I am more and more considering a handle-based scheme for the hnix-store API.

jwiegley:
It's a fair bit going on, but I like it. What do you think @ryantrinkle?

Ericson2314:
> Pending questions to be resolved later on, in other PRs

These things do further my suspicion that it's time to move to a monorepo so it's easier to experiment without how to integrate the various libraries while keeping things modular.

jwiegley:
I'm generally a fan of monorepos.

layus:
Short update about performance: 
<pre>
$ NIX_DATA_DIR=$PWD/hnix/data command time ./dist-newstyle/build/x86_64-linux/ghc-8.10.1/hnix-0.10.1/x/hnix/build/hnix/hnix --eval -E '"${(import <nixpkgs> { }).hello}"'
"/nix/store/w9yy7v61ipb5rx6i35zq1mvc2iqfmps1-hello-2.10"
<strong>22.21user</strong> 0.67system 0:22.95elapsed 99%CPU (0avgtext+0avgdata 1615464maxresident)k
0inputs+0outputs (0major+400263minor)pagefaults 0swaps
$ command time nix-instantiate --eval -E '"${(import <nixpkgs> {}).hello}"'                      (nix)
"/nix/store/w9yy7v61ipb5rx6i35zq1mvc2iqfmps1-hello-2.10"
<strong>0.22user</strong> 0.03system 0:00.25elapsed 100%CPU (0avgtext+0avgdata 111252maxresident)k
0inputs+0outputs (0major+25171minor)pagefaults 0swaps
</pre>

22 each time :-). And yet 100 times slower. Good news is that we have some room for improvement :joy:

jwiegley:
@layus Parsing and evaluating are each still very slow.

layus:
@jwiegley Any known culprit for that ?

layus:
Fixed a subtle bug where addToStore never added paths recursively... Took me a few hours, so I had to share this one ;-)

```diff
 instance (MonadIO m) => MonadStore (StandardTF r m) where
-  addToStore a b d c = StandardTF $ lift $ lift $ addToStore a b c d
+  addToStore a b c d = StandardTF $ lift $ lift $ addToStore a b c d
   addTextToStore a b c d = StandardTF $ lift $ lift $ addTextToStore a b c d
```

bqv:
So this is where you're hiding all the cool new stuff!

layus:
> So this is where you're hiding all the cool new stuff!

Yes indeed. And I would love to get feedback on it :-). /ping @expipiplus1 @jwiegley @Ericson2314 @kmicklas  :-)

It is far from perfect, but the decisions made here are not jeopardizing future improvements.
I am very willing to make changes upon request. It would be too sad if this bitrots (or brainrots, as I have tons of other stuff to do).

bqv:
@layus do you perhaps have a nix derivation to build this branch? I'm struggling to find the right versions of hnix-store to do so, but I'd love to have a play around with this while we wait for it to be merged!

layus:
@bqv Oh yes, that's definitely a pain point. I am currently working on top of sorki's hnix-overlay. 

The best I can provide is my fork of the project.

1. Checkout https://github.com/layus/hnix-overlay at branch derivationStrict
2. call nix-shell at the root of that checkout
3. You can now move to your hnix checkout and build this branch with cabal {configure,build,run}

Hope it helps.

bqv:
@layus Ah, does that work for you? Unfortunately cabal fails to find even the most basic deps under nix-shell, and if i try and recreate that env with a nix derivation, I see errors about MonadStoreT (which i can't even find anywhere!)

layus:
Oh dear, this was severely broken by https://github.com/haskell-nix/hnix-store/pull/61. It will take some work to get around that. Possibly we should make this a WIP pending on a final decision on how to lift/rewrite the store API.

layus:
@bqv Hi, I have resurrected my hnix-store branch before the nar streaming merge for you to test. If you are still interested.

  `nix-build https://github.com/layus/hnix-overlay/archive/derivationStrict.tar.gz -A hnix` now works, so that's already something.

Following the procedure above should also work now since all the branches have been updated.

layus:
Testing also requires a valid NIX_DATA_DIR. You can find one in hnix in the data/nix submodule, or in any nix install.

```
NIX_DATA_DIR=$HNIX_CHECKOUT/data hnix --eval -E '"${(import <nixpkgs> { }).hello}"'
```

building firefox does not work in this state because discrepancies between nix and hnix

expipiplus1:
> There are no tests, but nixpkgs.hello gives the same hash with nix and hnix. Achieving this was quite a feat 

Is it possible to add a test for this?

> the decisions made here are not jeopardizing future improvements.

Sounds like no bad thing to merge then! (pending compatibility with the changes in `hnix-store`)

> brainrots

What a useful word!

Is it a large amount of work to cope with the changes in hnix-store?

I'll try and give this a proper understanding this weekend.

layus:
> > There are no tests, but nixpkgs.hello gives the same hash with nix and hnix. Achieving this was quite a feat
> 
> Is it possible to add a test for this?

Certainly. It should be fairly trivial as we already have tests about evaluating the same nix expr in both nix and hnix. 

> > the decisions made here are not jeopardizing future improvements.
> 
> Sounds like no bad thing to merge then! (pending compatibility with the changes in `hnix-store`)

Well, as the recent events showed, it is not that simple. It requires the hnix-store to provide a monad transformer API, so as to keep the same socket open during all of hnix operation.

I would say that this statement mostly defines the changes made here up to (and including) "Cache hashDerivationModulo results ourselves".

After that, it requires significant changes in the hnix store. Now it seems that I should split this PR into safe and WIP parts :-)

> > brainrots
> 
> What a useful word!

:-)
 
> Is it a large amount of work to cope with the changes in hnix-store?

All the complexity stems from knitting the hnix and hnix-store together. If we drop "Thread hnix-store monad into the monad stack" (the last commit here) we get a properly functionning derivationStrict primop.
The issue being that it opens a new connection to the store socket on each nix API call.

I got a set of working changes for that, including "Thread hnix-store monad into the monad stack" in this PR and the modifications in hnix-store available at https://github.com/layus/hnix-store/commits/derivationStrict (specifically commit https://github.com/layus/hnix-store/commit/63fd63471bbda283ea790c780ec0805b9388c119)

This last set of changes was broken by the merge. I went further and lifted everything even more in https://github.com/haskell-nix/hnix-store/pull/72 and https://github.com/haskell-nix/hnix/pull/755 (always two there are :-D). It is the best I could do to integrate the nar streaming API with the need to filter streamed nars based in a functions that must itself reside in a MonadNix context. (a nix function).

The more I write about this, the more clear it appears that "Thread hnix-store monad into the monad stack" is not yet ready for merge. I just tested my code without it, and it runs at the same speed.
  I will drop this last patch from the PR, add tests and some closed nix expression to build it.
This makes the PR build with hnix-store master branch. \o/

> 
> I'll try and give this a proper understanding this weekend.

That would be awesome :-)

bqv:
(wow it worked, awesome. thanks @layus!)

layus:
@bqv Thanks for testing :-) :tada: 

@expipiplus1 In case you would like an easy way to test: `nix-build https://github.com/layus/hnix-overlay/archive/derivationStrict2.tar.gz -A hnix`

I force pushed this PR to remove the unnecessarily litigious commit, and added a proper test for nixpkgs.hello. nixpkgs.firefox is out of range because of #756.

There are a lot of small changes, but really only the three last commits contain the core of this implementation. The rest can be merged separately with #757, on which this PR is based.

layus:
I tried to make tests work, but I guess we will have to wait for a proper hnix-store 0.3.0.0 release for this branch to CI nicely. Pending on https://github.com/haskell-nix/hnix-store/issues/76

layus:
@expipiplus1 May I gently nudge you to give this PR a try ? :-)
It now depends on a proper release of hnix-store. I do not see anything more I could do without feedback.

@Anton-Latukha The CI is red all over the place. Mostly thanks to cryptohash-sha256.which is incompatible with ghc 8.10. I was surprised that our CI does not rely exclusively rely on nix for proper dependency management. Do you think there is a way to coax cryptohash-sha256 into our system ? We could possibly vendor it in hnix-store-core. It is merely a single-file ffi bindings.

default.nix
```diff
@@ -223,6 +222,17 @@ let
...
+      # recent versions of base, for seemingly no valid reason.
+      # As the update is slow to happen, just jailbreak here
+      # See https://github.com/haskell-hvr/cryptohash-sha512/pull/3
+      # and https://github.com/haskell-hvr/cryptohash-sha512/pull/5
+      cryptohash-sha512 = pkgs.haskell.lib.unmarkBroken ( pkgs.haskell.lib.doJailbreak super.cryptohash-sha512 );
+
+      # 2020-12-07 hnix-store-remote fails when trying to connect to a real hnix daemon.
+      # probably due to nix sandbox restrictions.
+      hnix-store-remote = pkgs.haskell.lib.dontCheck super.hnix-store-remote;
+
```
expipiplus1, line 237: Just a side note, have you made PRs for these to NixOS/nixpkgs?

src/Nix/Effects.hs
```diff
@@ -226,36 +235,54 @@ print = putStrLn . show
...
 
-    -- | Add a file with the given name and contents to the nix store
-    toFile_' :: FilePath -> String -> m (Either ErrorCall StorePath)
+    -- | Add a path to the store, with bells and whistles
+    addToStore :: StorePathName -> FilePath -> RecursiveFlag -> RepairFlag -> m (Either ErrorCall StorePath)
+    default addToStore :: (MonadTrans t, MonadStore m', m ~ t m') => StorePathName -> FilePath -> RecursiveFlag -> RepairFlag -> m (Either ErrorCall StorePath)
+    addToStore a b c d = lift $ addToStore a b c d
+
+    -- | Add a nar (action) to the store
+    -- addToStore' :: StorePathName -> IO Nar -> RecursiveFlag -> RepairFlag -> m (Either ErrorCall StorePath)
```
expipiplus1: remove?

src/Nix/Effects.hs
```diff
@@ -226,36 +235,54 @@ print = putStrLn . show
...
-  --TODO: Use a temp directory so we don't overwrite anything important
-  toFile_' filepath content = do
-    writeFile filepath content
-    storepath <- addPath' filepath
-    S.removeFile filepath
-    pure storepath
+  addToStore name path recursive repair = case Store.makeStorePathName name of
+    Left err -> return $ Left $ ErrorCall $ "String '" ++ show name ++ "' is not a valid path name: " ++ err
+    Right pathName -> do
+      -- TODO: redesign the filter parameter
```
expipiplus1, line 270: Can add an issue for this?

src/Nix/Effects.hs
```diff
@@ -226,36 +235,54 @@ print = putStrLn . show
...
+      res <- Store.runStore $ Store.addToStore @'Store.SHA256 pathName path recursive (const False) repair
+      parseStoreResult "addToStore" res >>= \case
+        Left err -> return $ Left err
+        Right storePath -> return $ Right $ StorePath $ T.unpack $ T.decodeUtf8 $ Store.storePathToRawFilePath storePath
```
expipiplus1, line 274: I guess it's no problem for nix store paths, but I always get the heebie-jeebies seeing encoding/decoding near file names.

src/Nix/Effects/Basic.hs
```diff
@@ -235,13 +225,13 @@ findPathM = findPathBy existingPath
     pure $ if exists then Just apath else Nothing
 
 defaultImportPath
-  :: (MonadNix e t f m, MonadState (HashMap FilePath NExprLoc) m)
+  :: (MonadNix e t f m, MonadState (HashMap FilePath NExprLoc, b) m)
```
expipiplus1, line 228: Shame that Haskell/mtl make the `b` necessary here, I guess it's just to make it fit a hole somewhere else.

src/Nix/Reduce.hs
```diff
@@ -66,19 +67,19 @@ import           System.FilePath
 
 newtype Reducer m a = Reducer
     { runReducer :: ReaderT (Maybe FilePath, Scopes (Reducer m) NExprLoc)
-                           (StateT (HashMap FilePath NExprLoc) m) a }
+                           (StateT (HashMap FilePath NExprLoc, MS.HashMap Text Text) m) a }
```
expipiplus1, line 70: Is it possible to give this Text, Text Map a type synonym (or newtype even if it's no trouble) to make it a bit more meaningful?

src/Nix/Effects/Derivation.hs
```diff
@@ -0,0 +1,395 @@
...
+import           Text.Megaparsec
+import           Text.Megaparsec.Char
+
+
+data Derivation = Derivation
+  { name :: Text
+  , outputs :: Map Text Text
+  , inputs :: (Set Text, Map Text [Text])
+  , platform :: Text
+  , builder :: Text -- should be typed as a store path
```
expipiplus1, line 60: Can add an issue for this TODO? (Or implement it I guess)

src/Nix/Effects/Derivation.hs
```diff
@@ -0,0 +1,395 @@
...
+  , args :: [ Text ]
+  , env :: Map Text Text
```
expipiplus1, line 62: Text is probably fine for most of these, but the env one does feel like it should be a bytestring. 

would it be possible to switch these over to bytestring (or short bytestring)? Or better yet newtypes around bytestring. 

Perhaps nix specifies that `getEnv` only works with well-encoded environment vars though, in which case I guess this is fine.

src/Nix/Effects/Derivation.hs
```diff
@@ -0,0 +1,395 @@
...
+  parens p = (string "(") *> sepBy p (string ",") <* (string ")")
+  list   p = (string "[") *> sepBy p (string ",") <* (string "]")
+
+  parseFixed :: [(Text, Text, Text, Text)] -> (Maybe Store.SomeNamedDigest, HashMode)
+  parseFixed fullOutputs = case fullOutputs of
+    [("out", _path, rht, hash)] | rht /= "" && hash /= "" ->
+      let (hashType, hashMode) = case Text.splitOn ":" rht of
+            ["r", ht] -> (ht, Recursive)
+            [ht] ->      (ht, Flat)
+            _ -> undefined -- What ?! -- TODO: Throw a proper error
```
expipiplus1: I guess `error "impossible: empty text in parseFixed"` would be slightly better, but no big deal either way

src/Nix/Effects/Derivation.hs
```diff
@@ -0,0 +1,395 @@
...
+      in case Store.mkNamedDigest hashType hash of
+        Right digest -> (Just digest, hashMode)
+        Left _err -> undefined -- TODO: Raise a proper parse error.
```
expipiplus1: Not sure on this one, should probably be at least `error ".."`

src/Nix/Effects/Derivation.hs
```diff
@@ -0,0 +1,395 @@
...
+          { inputs
+          , outputs = outputs'
+          , env = if useJson drv then env drv else Map.union outputs' (env drv)
+          }
+
+    drvPath <- pathToText <$> writeDerivation drv'
+
+    -- Memoize here, as it may be our last chance in case of readonly stores.
+    drvHash <- Store.encodeBase16 <$> hashDerivationModulo drv'
+    modify (\(a, b) -> (a, MS.insert drvPath drvHash b))
```
expipiplus1, line 261: This goes for all these modifies, the tuple has added a layer of laziness. Have you checked that this is sufficiently strict to not be a performance problem?

src/Nix/Effects/Derivation.hs
```diff
@@ -0,0 +1,395 @@
...
+
+    let outputsWithContext = Map.mapWithKey (\out path -> principledMakeNixStringWithSingletonContext path (StringContext drvPath (DerivationOutput out))) (outputs drv')
+        drvPathWithContext = principledMakeNixStringWithSingletonContext drvPath (StringContext drvPath AllOutputs)
+        attrSet = M.map nvStr $ M.fromList $ ("drvPath", drvPathWithContext): Map.toList outputsWithContext
+    -- TODO: Add location information for all the entries.
```
expipiplus1, line 266: can add issue?

src/Nix/Effects/Derivation.hs
```diff
@@ -0,0 +1,395 @@
...
+      return $ pathToText $ Store.makeStorePath "/nix/store" ("output:" <> Text.encodeUtf8 o) h name
+
+    toStorePaths ctx = foldl (flip addToInputs) (Set.empty, Map.empty) ctx
+    addToInputs (StringContext path kind) = case kind of
+      DirectPath -> first (Set.insert path)
+      DerivationOutput o -> second (Map.insertWith (++) path [o])
+      AllOutputs ->
+        -- TODO: recursive lookup. See prim_derivationStrict
+        -- XXX: When is this really used ?
+        undefined
```
expipiplus1: can change to `error`?

src/Nix/Effects/Derivation.hs
```diff
@@ -0,0 +1,395 @@
...
+      builder     <- getAttr   "builder"                   $ extractNixString
+      platform    <- getAttr   "system"                    $ extractNoCtx >=> assertNonNull
+      mHash       <- getAttrOr "outputHash"        Nothing $ extractNoCtx >=> (return . Just)
+      hashMode    <- getAttrOr "outputHashMode"    Flat    $ extractNoCtx >=> parseHashMode
+      outputs     <- getAttrOr "outputs"           ["out"] $ mapM (fromValue' >=> extractNoCtx)
+
+      mFixedOutput <- case mHash of
+        Nothing -> return Nothing
+        Just hash -> do
+          when (outputs /= ["out"]) $ lift $ throwError $ ErrorCall $ "Multiple outputs are not supported for fixed-output derivations"
```
expipiplus1, line 310: I don't know on the nix side of things, but could this work with a single output not named "out"?

expipiplus1 (approved):
Looks great, happy to merge! Any of the nits I mentioned could be done now or have issues created and done at a later time.

It would be good to have some more tests for the various improvements here, but that could also be done later.

default.nix
layus: I did not, but now it's done :-)

- cryptohash-sha512:  https://github.com/NixOS/nixpkgs/pull/106333
- hnix-store-remote: https://github.com/haskell-nix/hnix-store/issues/80

src/Nix/Effects.hs
layus: I have a WIP PR. Not really an issue though: https://github.com/haskell-nix/hnix/pull/755

src/Nix/Effects.hs
layus: Yes, maybe make a further issue for it. As long as hnix-store uses ByteStrings and hnix FilePaths, this will remain a pain. That being said, If this could be the only place where the conversion happens, it remains quite clean. At the moment such conversions also happen here and there in Effects/Derivation.hs.

layus: I guess https://github.com/haskell-nix/hnix/issues/472 is the right tracking issue.

src/Nix/Effects.hs
layus: Yep, there is definitely something to do about this. expipiplus1 is also quite bothered with that (https://github.com/haskell-nix/hnix/pull/554/files#r538011509).

I guess https://github.com/haskell-nix/hnix/issues/472 is the right tracking issue.

src/Nix/Effects/Basic.hs
layus: We could move to a more involved type. Possibly something like `Has ExprCache` and `Has DrvHashCache` but for StateT instead of ReaderT. Looks a lot lensy to me. I did not want to bother generalizing too much because

1. This might move to hnix-store if we end-up weaving the effect stacks
2. There does not seem to be a need for more stateful values
3. Usage of this state is limited to a few fucntions.

sorki (approved):
LGTM!

I would love to see this use `System.Nix.Derivation` based on `nix-derivation` but since it doesn't map 1:1 and possibly needs another wrapper we can poke that later.

Nice work!

src/Nix/Reduce.hs
layus: Again, this is supposed to be a short-lived state addition, but we could certainly do that.

src/Nix/Effects/Derivation.hs
layus: What we want here is a type that carries the proof that `builder` is a valid path in the store. A `StorePath` does not immediately provide that. A `Store.StorePath` (a `StorePath` from hnix-store) does, but then we mix and match more than we would like to. Will keep as a TODO until we decide on migrating this code to the store, or at least unify the Derivation datatype.

src/Nix/Effects/Derivation.hs
layus: `env` in this context is the set of environment variables available to the builder (`buildCommand`, `cmakeFlags` and the like). I am not sure how these are passed, but as they come from nix files that are parsed as text, having them as text does not cause any loss of precision, right ? Something else to ponder when unifying Dereivation datatype with the store.

src/Nix/Effects/Derivation.hs
layus: Yes. What I wanted is to make the parser fail with a parse error. But `error` will do for now :-)

src/Nix/Effects/Derivation.hs
layus: No, I did not and I am not aware of any way to do so. I will provide a quick, strict datatype so as to address this concern and naming as stated above.

src/Nix/Effects/Derivation.hs
layus: Well, I am not too sure what the location should be. I gues the right location is the builtin invocation, which is not very relevant as there is usually only one builtin.derivationStrict, and that is inside builtin.derivation.

Also, I am not sure how to access that location.

src/Nix/Effects/Derivation.hs
layus: At the moment the nix daemon will refuse fixed-outptut derivations that have custom outputs, including a single one not named "out". We just fail earlier.

src/Nix/Effects/Basic.hs
expipiplus1: Oh sure, I was just musing :)

layus (commented):
Added one commits to address all the received comments.

I tried to make the eval state typed, but failed halfway. I am not even sure what I am manipulating. It looks like all the types align magically, but that there might be no centralized place to define them. It creates circular inclusions.
Looks like a nontrivial refactor is needed for that. See https://github.com/layus/hnix/commit/63ab545675e00eb58b867a405037f95fc59a6c76 for my experiment.

Anton-Latukha:
About https://github.com/haskell-nix/hnix/pull/554#issuecomment-740180868:
> @Anton-Latukha The CI is red all over the place. Mostly thanks to cryptohash-sha256.which is incompatible with ghc 8.10. I was surprised that our CI does not rely exclusively rely on nix for proper dependency management. Do you think there is a way to coax cryptohash-sha256 into our system ? We could possibly vendor it in hnix-store-core. It is merely a single-file ffi bindings.

It is a multipart question.
1. `I was surprised that our CI does not rely exclusively rely on Nix for proper dependency management.`
  It does not rely exclusively, nor should it. Initially, CI was testing only Nix build, and I also thought that is "right", but then - HNix is as Haskell project first, especially for releasing/sharing/working on it and for the build systems. That is why Cabal builds passes are mandatory and Nix builds are optional, Haskell projects are developed using Hackage&Cabal first, and Nix niceties and bootstrapping is suplemmentary.

2. `cryptohash-sha512` not supporting `base 4.14`:
 `hnix-store` uses it. No wonder HNix CI is red all over the place, [`hnix-store-core` Hackage Matrix also](https://matrix.hackage.haskell.org/#/package/hnix-store-core), project has no CI testing, and I just so far not merged CI into HNix-store, because HNix-store has a unique custom two-project Nix bootstrapping, which requires rethinking [haskell-nix/haskell-with-nixpkgs](https://github.com/haskell-nix/haskell-with-nixpkgs) or customizing everything to the HNix-store custom infrastructure.
  Seen that you sent a jailbreak patch to Nixpkgs Haskell, but what to do with a Cabal is an open question.
  Since there are HNix-store questions - logically - moving the questions/tasks there, would do backlog clean-up here and then there.

layus:
@Anton-Latukha Thanks for the clarifications. I was a bit sad to make CI go red with this PR. I guess we can live with that between releases.

Anton-Latukha:
Yep, I think we should make a release ASAP, and tackle this and then make a new one.

We can make a release tomorrow (today by my local time). And then approach this situation, and as soon as we complete it - release a new release. Since current `master` is `minor` release and derivationStrict introduces `major`, the proper solution would be to take some time, I graphed the current situation and this situation demands sorting things out with `cryptohash-sha512` upstream merger/transfer/fork, or working it around both in Cabal, Store config, Nixpkgs (which you send-out) - it is a coupling between 3 projects which should be resolved.

BTW - it is useless to override `cryptohash-sha512` in HNix `default.nix` - it is not interpreters closure - package should be overloaded inside the Store closure, HNix `default.nix` for has a hatch for that between projects:
```
  #  2020-05-23: NOTE: Currently HNix-store needs no overlay
  # hnix-store-src = pkgs.fetchFromGitHub {
  #   owner = "haskell-nix";
  #   repo = "hnix-store";
  #   rev = "0.2.0.0";
  #   sha256 = "1qf5rn43d46vgqqgmwqdkjh78rfg6bcp4kypq3z7mx46sdpzvb78";
  # };

  overlay = pkgs.lib.foldr pkgs.lib.composeExtensions (_: _: {}) [
    # (import "${hnix-store-src}/overlay.nix")
```
, the purpose of it - is that it loads the Store's Nix configuration if Store uses some overloads. So it travels back - for maintaining the configuration there - Store needs to have CI in the first place to show that configuration needs an override.

I would look at this tomorrow and forward, and would start solving particular tasks by opening reports in the Store, if I lag behind - anyone can move forward with any part of this.

Anton-Latukha:
Thing with `cryptohash-512` is simple - it got deprecated long ago (5y, actually); https://github.com/haskell-nix/hnix-store/issues/83 - that's all.

sorki:
And as usual, things are not that simple as they look https://github.com/dhall-lang/dhall-haskell/pull/1822#issuecomment-636333155

Btw no matter how many times you edit your message, the version sent over email is the first version.

Anton-Latukha:
The minimal observable plan:

1. Merge https://github.com/haskell-nix/hnix-store/pull/84 to fix the HNix-Store build for GHCs < 8.10.
2. Make a working (semi-) HNix-Store fix release.
3. Which with already merged https://github.com/NixOS/nixpkgs/pull/106333 would allow Store to build in Nix-build in all cases, even with GHC 8.10.
4. This would allow merging that in large working HNix-Store release with `derivaitonStrict` PR here.
5. ... (further work)
6. Profit.

Overall this can go on in no particular order.

In the Store migration from Cryptohash to Cryptonite is, of course, possible, yesterday I tried to - migration essentially means by adding the lib rooting-out/throwing-out/rewriting the `Hash.hs` in the Store. Which is a pretty small sizable effort. Maybe someone with higher skill would find even a smaller way to migrate or who knows which interface is really needed. Current `Hash.hs` essentially duplicates the library code and adds a possibility of handling the Truncated hash. Upstreaming into the library the Truncated hash mode would allow to make `Hash.hs` almost/nonexistent, but also truncation of course can happen on top of hashes.

Overall we can merge it now and supply fixes on top, but also with small effort of merging the minimal fix, making working Store release and using it here - would allow to merge the work properly.

Anton-Latukha:
I shipped everything possible to `hnix-store` to make a proper working release, including temporary dep fix https://github.com/haskell-nix/hnix-store/pull/94.

Now waiting until someone who has access to Hackage comes along and makes a release.

bqv:
Is this mergeable now? Or does it strictly need a new -store release?

Anton-Latukha:
This PR just got in the middle of a situation, it uses code that is in `0.3` release that is broken and needs a fix. This PR "as is" merges the broken `hnix-store` release into the `hnix`. It is not a big deal, if the `hnix-store` project does not need to work, and if `hnix` does not need to work further :wink:.

Long story short - we currently have a `hnix-store` Hackage bottleneck, it seems there are only 2 people with access to the Hackage for it, and I guess only `shea` with admin access rights, and they did not respond for the last several weeks.

We can merge this on the trust and break `master` and all adjust to the problem. Any of us can merge this any time. But in any case we would anyway require for the `hnix-store` situation to resolve. We can patch Cabal dep by `cabal.project` and Nix config through overlay, and then remove all that when `hnix-store` releases work. But we also can just address the problem directly and resolve `hnix-store` processes bottleneck that was present there for some time and resolve all of the processes to the normal stable state.

I already went above and beyond in the HNix-store, fixed and prepared everything. What is needed is to publish `master` as a release with Hackage access and also revision fix the previous two releases.

I also want to merge this, and make a release right then, but with the projects that work for the contributors and general public.

Someone with admin level of HNix-store Hackage access needs to resolve the knot, or in a couple of weeks more, we would arrive at republishing the own group project to Hackage under a different name just to bring it and the processes into the working state.

bqv:
Since the hnix-store release is a hackage issue, not a github one, does it not make sense to merge this and deal with fiddling with package versions in another PR? That issue is entirely unrelated to this PR's goal, and in that sense this PR is complete (it's in fact even usable as long as you build it all yourself)

expipiplus1:
I am very much in favour of merging now (with an override in cabal.project

I think it would also be reasonable to contact the Hackage trustees if the
hnix store package is abandoned and the maintainers uncontactable.

Anton-Latukha:
For me, from the call they were unresponsive for a little more than 2 weeks, which is a requirement for me. On NY moreover may still be understandable.
But it is painful, because that situation presses on several pain points at the same time, last two HNix-store-core and last two (and only) HNix-store-remote releases are broken, the `cryptohash` is abandoned and needs to be switched from ASAP, HNix overall majorly depends on HNix-Store state of affairs, which showed here and this pretty important PR got stuck waiting.
`Shea` and `Greg` are essentially authors of HNix-Store and essentially created it, so I personally can wait until the situation resolves normally.
So I currently would say that calling the Hackage trustees I see as too hard to soon of a measure. We just created a bazar here around the bouquet of issues discussing and drawing their attention to resolve it. I personally want to wait and give them the time of a couple of weeks more and try to reach them through different channels, maybe they would show up.

---

Nevertheless, after #782 #785 #786 (I plan to make it in a couple of hours) the mentioned deps would be riding high on fixes. And PR should be able to merge normally. And then we would have a window to sort things out.

Anton-Latukha:
Good news - `shea` came around and `sorki` got Hackage rights to the `hnix-store-core`, now the awaited releases of Store working versions to Hackage much more probably to happen.

Anton-Latukha:
I did what is possible down the stack to clean everything up.

In https://github.com/haskell-nix/hnix/pull/785 gathered current overrides, there is example how to do Cabal monorepo overrides.

Thou, it is saner and cleaner to wait on the Store releases and do things sequentially.
In Store I started Hash refactoring&preparation, upon Store release, this is a support of the changes: https://github.com/haskell-nix/hnix/pull/785/commits/61ec2b4677cef6eb182adbeab3ffc907c5d96db5, since it sitst on top of new module from this PR and I already uploaded it.

Thanks for everything.

Have a nice New Year.

bqv:
To be clear, there's no obstacle to this PR in particular being merged now, no? Hnix-store releases are now up to date, so any further refactoring should almost certainly be a separate, new PR

hnix.cabal
```diff
@@ -401,8 +402,9 @@ library
     , gitrev >= 1.1.0 && < 1.4
     , hashable >= 1.2.5 && < 1.4
     , hashing >= 0.1.0 && < 0.2
-    , hnix-store-core >= 0.1.0 && < 0.3
-    , http-client >= 0.5.14 && < 0.6 || >= 0.6.4 && < 0.8
+    , hnix-store-core >= 0.3.0 && < 0.4
```
bqv: @layus if you wouldn't mind updating these bounds for the latest release, perhaps @expipiplus1 would be happy to merge this?

bqv (changes requested):

Anton-Latukha:
@layus

I know that the bug PRs are hard to do and merge.

I personally did as much as possible to improve the situation around the PR to make a good merge.
PR is from a fork - I want to provide the actions that left directly, but can not push to it.

Since this is a big improvement and change - people would frequently refer back to the merge and inside of it.

The situation around and the PR should be update-rebaseable at the very least, and the merge situation would become better.

I [asked](https://github.com/haskell-nix/hnix-store/issues/85#issuecomment-752768921) `sorki` to notify the team when the Store releases would work, he seems to encounter some bug releasing the projects to Hackage. But I am sure that would be resolved.

Please, can you respond what from the:
a) Introduce the dependency package support as a PR, before the PR that relies on it.
b) Clean-up the overrides.
c) Drop the HNix-Store `0.3` boundary, if it is not needed.
d) Upgrade to HNix-Store `0.4` in the PR, it is something small like 61ec2b4.
e) Or somehow in another way make the `cabal build`s work in the CI.
f) Maybe you got tired and want to merge it "as is", then we can meet in the middle and merge it in that case, and, I promise, I would deal with the consequences and clean-up. According to CI on "as is" we would have a build breakage in the `master` history around this important merge, which some people may refer to in the future.

What you consider the right steps, what way we go further.

layus:
Hi Anton,

Thanks for the sustained work you are doing to untangle this Gordian knot.

I am willing to update this PR to make it perfectly working with the new store release. But any merge path is fine with me.

I think I left the PR open for edits. You are welcome to make progress before I am able to :-), probably next week.

In the mean time, happy new year to you all !

Le 30 décembre 2020 23:25:57 GMT+01:00, Anton Latukha <notifications@github.com> a écrit :

Anton-Latukha:
Reshuffled the history, cleaned it up, and minimized transitions as much as I could.

Figured-out and fixed a couple of things in config changes.

Now, it looks strangely familiar:

----
<details><summary>On macOS compilations fails due to non-existent IPC calls </summary>
<p>

```
command was: /nix/store/0nlwbj5qbdf7r58x2iqz0a17az2j5fdn-clang-wrapper-7.1.0/bin/clang -c dist/build/System/Linux/Namespaces_hsc_make.c -o dist/build/System/Linux/Namespaces_hsc_make.o -D__GLASGOW_HASKELL__=810 -Ddarwin_BUILD_OS=1 -Dx86_64_BUILD_ARCH=1 -Ddarwin_HOST_OS=1 -Dx86_64_HOST_ARCH=1 -I/nix/store/0jj05cyma647rywsqpq3darnzl6a520s-libc++-7.1.0/include -I/nix/store/ns7f13gldnk83l2fgjlzbx9bnfvc2g72-libc++abi-7.1.0/include -I/nix/store/6snmf1z43ay6m1b8jv8dr4h8im056a62-compiler-rt-7.1.0-dev/include -I/nix/store/q8xcp8hsa35wjki8mad0sbk4nzqrdkb5-libiconv-osx-10.12.6/include -I/nix/store/0jj05cyma647rywsqpq3darnzl6a520s-libc++-7.1.0/include -I/nix/store/ns7f13gldnk83l2fgjlzbx9bnfvc2g72-libc++abi-7.1.0/include -I/nix/store/6snmf1z43ay6m1b8jv8dr4h8im056a62-compiler-rt-7.1.0-dev/include -I/nix/store/q8xcp8hsa35wjki8mad0sbk4nzqrdkb5-libiconv-osx-10.12.6/include -Idist/build/autogen -Idist/build/global-autogen -include dist/build/autogen/cabal_macros.h -I/nix/store/lj34zl1kfjd2556aslrn8hsyn0ng1hzb-ghc-8.10.3/lib/ghc-8.10.3/unix-2.7.2.2/include -I/nix/store/lj34zl1kfjd2556aslrn8hsyn0ng1hzb-ghc-8.10.3/lib/ghc-8.10.3/time-1.9.3/include -I/nix/store/lj34zl1kfjd2556aslrn8hsyn0ng1hzb-ghc-8.10.3/lib/ghc-8.10.3/bytestring-0.10.12.0/include -I/nix/store/q8xcp8hsa35wjki8mad0sbk4nzqrdkb5-libiconv-osx-10.12.6/include -I/nix/store/lj34zl1kfjd2556aslrn8hsyn0ng1hzb-ghc-8.10.3/lib/ghc-8.10.3/base-4.14.1.0/include -I/nix/store/9m9q7bmn937d15p8q9m26fpxypi8vg92-gmp-6.2.1-dev/include -I/nix/store/lj34zl1kfjd2556aslrn8hsyn0ng1hzb-ghc-8.10.3/lib/ghc-8.10.3/integer-gmp-1.0.3.0/include -I/nix/store/lj34zl1kfjd2556aslrn8hsyn0ng1hzb-ghc-8.10.3/lib/ghc-8.10.3/include -I/nix/store/jikzfsnc21fsfzn991md0vi8j14p02sm-libffi-3.3-dev/include -I/nix/store/lj34zl1kfjd2556aslrn8hsyn0ng1hzb-ghc-8.10.3/lib/ghc-8.10.3/include/
error: Namespaces.hsc:79:16: error: use of undeclared identifier 'CLONE_NEWIPC'
    hsc_const (CLONE_NEWIPC);
               ^
Namespaces.hsc:79:16: error: use of undeclared identifier 'CLONE_NEWIPC'
Namespaces.hsc:79:16: error: use of undeclared identifier 'CLONE_NEWIPC'
Namespaces.hsc:80:16: error: use of undeclared identifier 'CLONE_NEWNET'
```

----
</details>

<details><summary>In Nix builds fail because they use network access in tests </summary>
<p>

```
  Eval comparison tests
    builtins.toJSON.nix:                                             FAIL
      Exception: Network.Socket.connect: <socket: 13>: does not exist (No such file or directory)
    builtins.lessThan-01.nix:                                        FAIL
      Exception: Network.Socket.connect: <socket: 14>: does not exist (No such file or directory)
    builtins.getContext.nix:                                         FAIL
      Exception: Network.Socket.connect: <socket: 13>: does not exist (No such file or directory)
    builtins.replaceStrings-01.nix:                                  FAIL
      Exception: Network.Socket.connect: <socket: 13>: does not exist (No such file or directory)
      
  Nix (upstream) language tests
    eval okay
      eval-okay-eq-derivations:                                      FAIL
        Exception: Network.Socket.connect: <socket: 13>: does not exist (No such file or directory)
      eval-okay-context:                                             FAIL
        Exception: Network.Socket.connect: <socket: 13>: does not exist (No such file or directory)
```
----
</details>

<details><summary>On GHC 8.10 GitHub decided no matter what PR is - to take a New Year break after 10 minutes of work </summary>
<p>

10 minutes...
```
Error: The operation was canceled.
```

----
</details>

<details><summary>And what really rings true is the Cabal build </summary>
<p>

```
   Eval comparison tests
    nixpkgs.hello.nix:                                               FAIL
      Exception: /nix/store/q5xckc93ra0fahpj6z01ffrxj3pa0qa8-source: openBinaryFile: does not exist (No such file or directory)
```

----
</details>

What would you do - every build is familiar in its own way. But only the last one is, as always, being true. If GitHub was not lazy today - the GHC 8.10 build would've told us the exact same story (since it tells it locally).

It compiles, but fails on that little test.

Test group defined as:
```haskell
genEvalCompareTests = do
    td <- D.listDirectory testDir

    let unmaskedFiles = filter ((==".nix") . takeExtension) td
    let files = unmaskedFiles \\ maskedFiles

    pure $ testGroup "Eval comparison tests" $ map (mkTestCase testDir) files
  where
    mkTestCase td f = testCase f $ assertEvalFileMatchesNix (td </> f)
```

What we essentially have is a 1 test to fix. And that is a test introduced here: `nixpkgs.hello.nix`([open](https://github.com/haskell-nix/hnix/blob/ac7450a3ba3b648b3738b3930ccb307b7795c562/tests/eval-compare/nixpkgs.hello.nix)):
https://github.com/haskell-nix/hnix/blob/ac7450a3ba3b648b3738b3930ccb307b7795c562/tests/eval-compare/nixpkgs.hello.nix#L1-L22

I would think and also wait on thoughts on what is the thing with this test.

Anton-Latukha:
This happens because:
```
fetch 'nix/fetchurl.nix` was not found in the Nix search path
```

Dolstra recently moves/reduces  the helper `.nix` files.
Recently _afair_ he reduced the `derivation.nix`, which we internally use.
In derivation it now lives in the `./share/nix/corepkgs/fetchurl.nix`.

Anton-Latukha:
It is somewhere between functions `fetchTarball`, it's `fetch` and `instantiateExpr`.

Anton-Latukha:
Why and how it looks into it.
`corepkgs/fetchurl.nix` needs inclusion into cabal file.
...
Maybe it is `foldNixPath`.

layus:
That's a lot of small issues bundled together. Let's try them one at a time:

---

```fetch 'nix/fetchurl.nix` was not found in the Nix search path``` comes is an issue I am unsure how to solve. I ran all my tests with 
`NIX_DATA_DIR=/path/to/hnix/data` env var. The tests expect to have it in the same location. (See https://github.com/layus/hnix/blob/2ec374d19e4a4090abed5513faf6ad6644f0ff8b/tests/Main.hs#L99 ). If it fails, it may be because we do not bundle these files. Or something else. Not quite sure.

For this PR, we need to bundle the files correctly so that `hnix` invocations can find them by default in the hnix installation.

As for the 

> Dolstra recently moves/reduces the helper .nix files.
> Recently afair he reduced the derivation.nix, which we internally use.
> In derivation it now lives in the ./share/nix/corepkgs/fetchurl.nix.

This can wait for a future PR. It is not part of the nix interface. It's implementation specific. We could also decide to bundle it in a single value just like we do for [`builtins.derivation`](https://github.com/layus/hnix/blob/2ec374d19e4a4090abed5513faf6ad6644f0ff8b/src/Nix/Builtins.hs#L172-L200)

---

In nix, the build fails because we need network access. That was already the case before. Worth to be fixed, but not necessarily here.

But they _also_ fail because the need a working nix daemon. In hnix-store @sorki  make a great work at both i) hiding these tests behind a flag and ii) wrapping everything in namespaces to provide a sandboxed nix daemon.

We could import either. And possibly both. So we would have flags tho disable tests that require networking, and provide a sandboxed nix daemon for tests.

It is a bit sad that the current implementation relies on the remote version of the hnix-store. a read-only version would fit the tests better. But it does not exist yet. Among other things because it would require some internal state that needs some architectural decisions about where/how to keep it. So for now we need the real nix daemon somewhere.

---

I guess that already some stuff to grind :-)

layus:
As for `<nix/fetchurl.nix>` it is called from nixpkgs @ https://github.com/NixOS/nixpkgs/blob/1dc37370c489b610f8b91d7fdd40633163ffbafd/pkgs/build-support/fetchurl/boot.nix#L11

The link references the same nixpkgs commit as the one we fetch with fetchTarball.

The solution is a proper NIX_DATA_DIR being set because that gets used as the `<nix/...>` prefix in the default nix path: https://github.com/layus/hnix/blob/2ec374d19e4a4090abed5513faf6ad6644f0ff8b/src/Nix/Builtins.hs#L313-L326

layus:
Or rather a preperly filled `getDataDir` which is generated automatically in Paths_hnix.hs and gets filled by `Data-Files:` cabal directive.

layus:
`builtins.fetchTarball` is broken because we have https://github.com/layus/hnix/blob/77859172494967a9de6292b4021844efe3b547ab/tests/Main.hs#L98 which downloads to $PWD while still printing a path in /nix/store/... . I wonder if it is a new nix feature. local?root= seems a hack that could be bo more supported in new nix builds.

layus:
Indeed. `local?root=...` is now `local?store=...` :dromedary_camel:

Anton-Latukha:
Great read. I never thought that it might be the Nixpkgs call of the file, automatically assumed it is or Haskell code or `nix-instantiate` invocations, so dug there.

I wanted to propose simple separation of consciences: cherry-pick the test into the separate PR and merge current. And deal with fixing the envocation in the test separately.

bqv:
> As for the 
> 
> > Dolstra recently moves/reduces the helper .nix files.
> > Recently afair he reduced the derivation.nix, which we internally use.
> > In derivation it now lives in the ./share/nix/corepkgs/fetchurl.nix.
> 
> This can wait for a future PR. It is not part of the nix interface. It's implementation specific. We could also decide to bundle it in a single value just like we do for [`builtins.derivation`](https://github.com/layus/hnix/blob/2ec374d19e4a4090abed5513faf6ad6644f0ff8b/src/Nix/Builtins.hs#L172-L200)
 

If they're needed, and it's not going to leave this PR in purgatory for *another* seven months, I think this is the most sensible option

layus:
> I wanted to propose simple separation of consciences: cherry-pick the test into the separate PR and merge current. And deal with fixing the invocation in the test separately.

Ok, let's go with that. HNix does not yet support chroot stores, but runs the tests with it. It worked while we were only using the real nix for real store operations. Now it breaks.

bqv:
Hooray, we might be merged before the new year!

Happy holidays, friends.

layus:
@Anton-Latukha I think these two new changes are enough, even if not entirely satisfactory. It would be fun to merge this in late 2020. But getting the first merge of 2021 would be an achievement too.

I would appreciate if you could merge it yourself after running the checks you consider necessary :-D.

Cheers ! And happy new year :-)

Anton-Latukha:
Yep, I'd already cleaned-up and merged it as https://github.com/haskell-nix/hnix/pull/554#issuecomment-753012228 when I wrote it, but I actually wrote that comment from the `TERM==linux`, I had a hard a severe rare case of unsolvable by me hardware problem because of Linux Kerne team doing the right things with NVidia's "GPL condom" and those overall. I made test&code assessment in the Linux console, Emacs is a terminal OS, that saves the ball every time.
Now everything works fine, just a minute.

Anton-Latukha:
I formed that test into a separate PR and dropped its revert here.

Cabal builds already worked - which is a minimal and enough requirement for the PR merges for me. There is no sense in breaking the `master` for everyone because some of us want to merge the patch that can not under the most basic conditions.

---

I am not an authority, but I recommend to think about our pathways - all our work and infrastructure, Nix & Nixpkgs Haskell, Stack also - are still, in the end, depend and use Hackage and its Cabal descriptions, and so - our Hackage community releases which are Cabal builds and descriptions. That is why it is the main thing for me - to keep Cabal builds working. On top of it - that keeps the history and the code open for any person in the community, it is a minimal requirement and enough to do Haskell work.

If I ever merge my or other PRs that had the Cabal builds broken but I merged them - you can beat my hands with a bamboo stick.

---

I wish one day to be as productive as you, @layus

bqv:
Thank you all!

Anton-Latukha:
That is not all, do not put a cart before a horse.

We have a big API. And a big API change.
For the release and to respect people that use HNix - the changelog needs to be populated first-class. Putting in the short human form the cliffnotes to guide fellows what happened from what they may need to be guided in a human for to update to new HNix. There is a big list of changes to be parsed and simply explained in the changelog. It is late and I would not finish it today.
If our API is as big and open - we would need to do this guidance all the time for every function we add, and for any function type change - that is additionally also a major release breakage level. Since almost all of them are made public - so we must support what we provide and promise to support to the public.
That is also why I want for us to do an API clean at some point. We need to expose only what downstream needs, and put a disclaimer that if something is additionally needed - for people to open reports or PRs opening it, and issuing those minor releases fast.
The changelog effort and a number of breaking major releases are linear to the API size of the project.
Changelog is populated in: https://github.com/haskell-nix/hnix/pull/788

PR #563, opened 2020-05-20
dep: allow pretty-show 1.12

sorki:
Allows building with GHC 8.10.

### DIFF
--- a/hnix.cabal
+++ b/hnix.cabal
@@ -915,7 +915,7 @@ library
       , cryptohash-sha256 >= 0.11.101 && < 0.12
       , cryptohash-sha512 >= 0.11.100 && < 0.12
       , haskeline >= 0.7.4.2 && < 0.8
-      , pretty-show >= 1.9.5 && < 1.11
+      , pretty-show >= 1.9.5 && < 1.12
       , serialise >= 0.2.1 && < 0.3
   -- if !flag(profiling)
   --   build-depends:


Anton-Latukha:
I know for you who done the work - it looks logical.

Please state more explicitly. There is pretty implicit big gap in logic between the `pretty-show < 0.12` and the building with `GHC 8.10`.

We would only save our time, people would not deduct, we would save back'n'forward effort, and you would not need to switch tasks and put more effort.

There is no mentions of the package in: GHC 8.10 migration docs https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/8.10.

There is no info on the product page: https://hackage.haskell.org/package/pretty-show
Also they deleted their Wiki.

There is `pretty-show < 0.12`, but Hackage has only `pretty-show 1.10`, so from where the `0.11` can come from?

Also look at the semantics meaning of the Report name.

---

Please provide a more instructive explanation.

<sub> I also probably should address in the names not "the package" but more properly - the constraint.
Since I follow my own naming protocol (Google-style like) of the naming the commits - so I use `{add,upd,rm,mv}` descriptions pretty constantly.

sorki:
Yes, it's actually an oversight it looks like.

I was going to write `ssia` into the description as I think this is obvious in `nixpkgs` context but to clarify building with `GHC 8.10` means using `pkgs.haskell.packages.ghc8101` and the set of packages it currently contains, I can add `nixpkgs` commit for the reference next time.

Since there are no guidelines you can't expect people to follow them so please propose them explicitly.

> Also look at the semantics meaning of the Report name.

I don't know what you mean by this.

Anton-Latukha:
The name of the PR is "dep: allow pretty-show 1.12", but the PR is `pretty-show < 1.12`, so it can not "allow pretty-show 1.12" in the first place. And secondly - that even `1.11` version not even exist.

And my further comment is about me tending to make the same `<` mistake.

I just tried to be very polite by the nodding into it.

sorki:
Thanks, it was a good catch - tbh I don't find anything of that impolite but a bit of unnecessary - I think we can get to the state where manual interventions (and errors caused by them) due to updating bounds become less of an issue thanks to the automation and CI.

I still think written guidelines are a good idea (especially branching / release management) wise, will open an issue for that as well.

Issue #570, opened 2020-05-21
Create a badge that tracks project dependencies being satisfied


Anton-Latukha:
There is `packdeps` web service and the tool.

I prefer "the serverless" approach - the web service already provides the needed info. But it does not respec the preferred versions of the Hackage. That resulted into constant `unmet` because of `binary` deprecations. I opened an according issue to solve this in `packdeps`.

Tool has an option respect of the preferred versions.

PR #573, opened 2020-05-21
README: add dependencies status badge monitor

Anton-Latukha:
Solves Repor idea: https://github.com/haskell-nix/hnix/issues/570

Now we would be able to observe if there are new unmet dependency versions.

It also links to the detailed table on `packdeps`.

M  README.md


sorki (approved):
LGTM!

PR #579, opened 2020-05-23
default.nix: clean-up old overrides

### LINKED ISSUE #566: Refactor `default.nix`
It does contain a lot of superfluous things

- commented out `packageOverride` for `nix` (Anyone knows the reason for that?)
- overrides for `brittany` and few others that are not needed at all for `GHC8.6.5+`
- pinning of `nixkpgs` should not be the default as it possibly forces another GHC and its package set on you, better if done from something like `release.nix`.

Regarding `brittany` if it needs overrides it would be better to import it and related overrides
from separate file or drop it altogether as the devs who use it probably have it in their env already.

My proposal is to reduce it to minimum viable `default.nix` and `overlay.nix` for *temporary* package overrides, which should also carry the note why the override is needed.

Anton-Latukha:
Report: https://github.com/haskell-nix/hnix/issues/566

  * I'm compiling with 8.6.5 (in explanations that was for >= 8.6.3) - so `{aeson,base-compat-batteries}` packages no longer needs explicit contravariant.

  * Upstream type issues of `Diff` may be long solved already. So lets properly check Diff. 8.4.4 [does not compile from HNix `0.5`](https://matrix.hackage.haskell.org/package/hnix) and Diff is definitely not a cause of those.

  * All other no longer needed overrides. They were not documented why they were introduced, so it is impossible to check their reasons. Freed HNix from those overrides. So we always can reintroduce them with proper documentation.

All tests pass fine.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -82,23 +82,6 @@ let
       multistate        = doJailbreak (overrideCabal super.multistate (attrs: { broken = false; }));
       butcher           = doJailbreak (overrideCabal super.butcher (attrs: { broken = false; }));
 
-      brittany = doJailbreak (self.callCabal2nix "brittany"
-        (pkgs.fetchFromGitHub {
-           owner  = "lspitzner";
-           repo   = "brittany";
-           rev    = "af227a797d588eda936280dc1c3b0b376735335e";
-           sha256 = "0l1nk4dgmlv8vl1d993vnyw3da0kzg4gq8c2zd8sd224f2rz6f35";
-           # date = 2019-12-20T01:20:07+01:00;
-         }) {});
-
-      ghc-exactprint = dontCheck (self.callCabal2nix "ghc-exactprint"
-        (pkgs.fetchFromGitHub {
-           owner  = "alanz";
-           repo   = "ghc-exactprint";
-           rev    = "91f54d7a7a1d8d2131c5e83d13dee6c9e8b57831";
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -71,40 +71,8 @@ let
...
-      # These packages only depend on contravariant when ghc >= 8.6.3
-      # Without normalizing the dependencies, our build fails with
-      # aeson and base-compat-batteries unable to find `contravariant`
-      aeson                 = addBuildDepend super.aeson self.contravariant;
-      base-compat-batteries = addBuildDepend super.base-compat-batteries self.contravariant;
-
-      mono-traversable  = dontCheck super.mono-traversable;
-      regex-tdfa-text   = doJailbreak super.regex-tdfa-text;
-      these             = doJailbreak super.these;
       semialign         = super.semialign_1_1;
```
sorki, line 74: Do you remember why this one is needed? I think I didn't need it for `GHC-8.8.1` nor `8.10.1` but haven't tested on `8.6.5`. Just for the reference.

sorki (approved):
Thanks, this does look good to me. I will possibly PR more minification when we figure out why some stuff is in place.

default.nix
Anton-Latukha: IDK/remember what was the message, not paid attention into it, just remember `8.6.5` was asking for it, so I left it.
First cleaning-up from all easy no longer needed stuff, so we would know what is left is indeed needed for some reason.

PR #594, opened 2020-05-27
get pinned nixpkgs with fetchTarball and guard with hash

ghost:
fetchGit is extremely slow with such a large repository, and fetchTarball is equivalent here.
In general it is good practice to have even builtin derivations guarded by hashes.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -10,12 +10,14 @@
 
 , rev ? "29d57de30101b51b016310ee51c2c4ec762f88db" #  2020-05-23: NOTE: UTC 17:00
 
+, sha256 ? "1wjljkffb3gzdvpfc4v98mrhzack6k9i7860n8cf5nipyab6jbq9"
+
 , pkgs ?
     if builtins.compareVersions builtins.nixVersion "2.0" < 0
     then abort "hnix requires at least nix 2.0"
-    else import (builtins.fetchGit {
-           url = "https://github.com/NixOS/nixpkgs/";
-           inherit rev; }) {
+    else import (builtins.fetchTarball {
+           url = "https://github.com/NixOS/nixpkgs/archive/${rev}.tar.gz";
+           inherit sha256; }) {
       config.allowUnfree = true;
       config.allowBroken = true;
       # config.packageOverrides = pkgs: rec {


sorki (approved):
Thanks, looks good to me. I think I'll propose another PR which will split pinning into `release.nix` and leave `default.nix` using systems `<nixpkgs>` by default.

Anton-Latukha:
@sorki 

As I understood you approved this.

I also approve this.

I love fast dev loop & accept small atomic PRs.

Tarball indeed is an old and still best practice.

default.nix
```diff
@@ -10,12 +10,14 @@
 
 , rev ? "29d57de30101b51b016310ee51c2c4ec762f88db" #  2020-05-23: NOTE: UTC 17:00
 
+, sha256 ? "1wjljkffb3gzdvpfc4v98mrhzack6k9i7860n8cf5nipyab6jbq9"
```
sjakobi, line 13: Noob question: Where do I find the `sha256` for a particular nixpkgs revision?

ghost: There are two easy ways: Using `nix-prefetch-url - -unpack`, or trying to build with a nonexistent one (easily created by toggling the initial '0' or '1' of an existing one).

sjakobi: Thanks! :)

Anton-Latukha: Yep, but there is still that inhumane `sha256:` next to it.
...

The important thing to know that Nix ("somehow") prefers `sha` search over `rev`/`version` search. Version information concidered irrelevant meta information. If you update a version, and not updated a sha - Nix would successfully be happy to take that `sha` from CDN and nicely compile/package it, and package it in beautiful inconsistent version information.

That is why we type a new version and then, invalidate the `sha` to get the `sha` collision resolve to put the `sha`.

When there is invalid `sha`, Nix suddenly checks version information.

It is technically understandable why it is simpler to just prefer the `sha`.

ghost: The thing is, git hashes are opaque to Nix and you would not be able to design a system of similar complexity for which they wouldn't be. You may be interested in https://discourse.nixos.org/t/obsidian-systems-is-excited-to-bring-ipfs-support-to-nix/7375, as that should enable Nix to verify build outputs using IPLD.

Anton-Latukha: Yep, Nix craves for IPFS integration.

I do not see why to not just get the secure cache from the central store, fetch the package from wherever it would come, and check it.

PR #595, opened 2020-05-27
bump aeson to 1.5, and with it these to 1.1

ghost:
I saw in https://github.com/haskell-nix/hnix/issues/551#issuecomment-633953865 that a bump of aeson was wished before the 0.7.2 release would be cut, so here it is.
At a glance, there don't seem to be any relevant changes:
https://hackage.haskell.org/package/aeson-1.5.0.0/changelog
https://github.com/bos/aeson/compare/1.4.7.1-r1...1.5.0.0
But 1.5.1 requires a bump to these to avoid diamond dependencies.
Changes also seem irrelevant and just related to the change in aeson:
https://hackage.haskell.org/package/these-1.1/changelog
https://github.com/isomorphism/these/compare/v1...v1.1 (most changes here are in other packages in the same repository)

These changes compile fine and tests are successful with GHC 8.8.3.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -10,12 +10,14 @@
 
 , rev ? "29d57de30101b51b016310ee51c2c4ec762f88db" #  2020-05-23: NOTE: UTC 17:00
 
+, sha256 ? "1wjljkffb3gzdvpfc4v98mrhzack6k9i7860n8cf5nipyab6jbq9"
+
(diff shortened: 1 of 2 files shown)


Anton-Latukha (approved):
Thank you. I hope we make a release soon.

PR #605, opened 2020-05-31
dafault.nix: GHC 8.8.3 as default; provide an easy part of the haskell.lib API

Anton-Latukha:
This would also make according to options in the CI work.

`doStrict` is ancient deprecated Nixpkgs option name, that is not used in Nixpkgs, nor in our `default.nix`, it is empty option.

For introduction `haskell.lib` options in CI before here. Sorry, that it is backward, I relearn the design flaws of Nix and its tooling usage again and again over the years.

These options provided - would start to work in CI now.

The rest of the options I would send-in PRs further.

---

Descriptions work, `nix-shell` works, `nix-build` works, project compiles, tests, benchmarks, profiles, with them enabled.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -30,6 +30,8 @@ generateOptparseApplicativeCompletions=${generateOptparseApplicativeCompletions:
 allowInconsistentDependencies=${allowInconsistentDependencies:-'false'}
 ghcjsTmpLogFile=${ghcjsTmpLogFile:-'/tmp/ghcjsTmpLogFile.log'}
 ghcjsLogTailLength=${ghcjsLogTailLength:-'10000'}
+# NOTE: If key not provided (branch is not inside the central repo) - init CACHIX_SIGNING_KEY as empty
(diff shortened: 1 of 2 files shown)


sjakobi (dismissed):
I'll approve just so you can merge if you want, @Anton-Latukha. I haven't taken a closer look.

Thanks for all the work you do! :)

Anton-Latukha:
"I haven't taken a closer look." - It is very OK @sjakobi.

Everything around Nix is a pretty hard & messy, I do not expect from myself nor from others to understand it great. Moreover, you already roughly explained where you coming from.

If Nix lang at least had the ability of good support in IDEs, so people would've been able to see what is in the scopes, what is imported, and mainly - what types functions and values have - then it would've been easier to work with it. But there are no such features, so it is still a heuristic development that heavily relies on try and error until it works.

sjakobi (approved):

Issue #624, opened 2020-06-12
Please, Cachix now requires Nix >= 2.2, so to fix builds update the Nix in Travis images before installing cachix


Anton-Latukha:

Anton-Latukha:
Done.

PR #625, opened 2020-06-12
CI: Travis: add update of Nix packages

Anton-Latukha:
Report: #624 

This was due because Nix installer installs very old version of `Nix`, and launches `Nix-daemon`. So now when we need right after right after the install to update - `Nix-daemon` and `Nix` are incompatible after the update.

After Nix updating itself, it fails to operate and needs HACKs to its daemon files and then restart of its daemon so Nix-daemon matches and supports the Nix.

The solution is tweaked out to work of the information from the closed and gorgotten by God report thread.

Just maybe this caveat and how to solve it, should to be documented in the Nix changelogs. And maybe the installer should instal new major version right away in the first place, so users and machines do not need to jump through the hoops on every install and figure-out why everything falls on its face and that is a Nix installation bug and then relearn all the install and migration bugs from the last two years.

We installing Nix on every CI launch, I do not even, why we need to jump though all this just to be able to install&use Cachix.

For what it is it was a major hurdle.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -119,6 +119,11 @@ before_script:
   - sudo mount -o remount,exec,size=4G,mode=755 /run/user || true
   - sudo mkdir -p /etc/nix
   - echo "trusted-users = root $USER" | sudo tee -a /etc/nix/nix.conf
+  #
+  # NOTE: Update Nix and Nix packages, Cachix requires that
+  - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then nix upgrade-nix; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then sudo nix upgrade-nix; fi
+  #
   # NOTE: macOS service restart
   - sudo launchctl kickstart -k system/org.nixos.nix-daemon || true
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -125,7 +125,20 @@ before_script:
   - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then sudo nix upgrade-nix; fi
   #
   # NOTE: macOS service restart
-  - sudo launchctl kickstart -k system/org.nixos.nix-daemon || true
+  # HACK: Because Nix can not update its daemon itself between 2.0 -> 2.2:
+  # 1. https://github.com/NixOS/nix/issues/3125#issuecomment-539667327
+  # 2. https://github.com/NixOS/nix/issues/3125#issuecomment-539771768
+  # Since Nix on macOS has problems restarting the service:
+  # Force the real disable, get target file from the symlink, unlink symlink, copy file, and reload service
+  - |
+    if [ "${TRAVIS_OS_NAME}" = "osx" ]; then true && \
+    sudo launchctl unload /Library/LaunchDaemons/org.nixos.nix-daemon.plist && \
+    export NIX_SERVICE_FILE=$(readlink /Library/LaunchDaemons/org.nixos.nix-daemon.plist) && \
+    sudo unlink /Library/LaunchDaemons/org.nixos.nix-daemon.plist && \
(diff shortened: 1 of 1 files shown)


abathur:
Just wanted to let you know (in case it lets you simplify anything--I haven't looked closely at your setup) that Nix support is working on Travis again, and the default version is updated to 2.3.6, as of about an hour ago.

Anton-Latukha:
@abathur 

Yes, I was in the middle of refactoring CI when it got released.

They basically switched the installer to multiuser installer by default, but it needs a service reload after trusted-users update, without reload the user was rejected and so the builds were unboxed (unpure) and so the system env went into Nix inputs, which triggered the whole stack rebuilds.

I have a long history about those installers, and I am still salty. But I also see my arguments from the talk were accepted as true statements in the end, operating from root is preserved, and even more - explicitly stated, as also explicitness of the installers, actions under superuser are especially explicitly logged.

Anton-Latukha:
@abathur

Also I see the service restart or even hack for macOS is still needed: https://travis-ci.org/github/haskell-nix/hnix/builds/701405539

Anton-Latukha:
@abathur

From which project are you that you know that Nix in Travis CI works again?

Anton-Latukha:
Well, macOS falls even with all current workarounds. !Fun!, it basically fails to update itself having a superuser permissions.

https://travis-ci.org/github/haskell-nix/hnix/jobs/701412861

Anton-Latukha:
If Nix was feeling Ok without updates for 2 years worldwide, I think we can allow and Ok to roll on current Nix version on macOS until we migrate to other CI.

abathur:
> From which project are you that you know that Nix in Travis CI works again?

Just a member from the Nix community (and macOS user) stepping in to make sure Travis CI works again (because I use it myself).

I hesitate to try and debug your issue because I'm not sure I understand why your setup script tries to run nix-channel and nix-env with sudo (and why it needs to update?). I don't want to insult you by second-guessing everything.

Anton-Latukha:
Because before this, not so long ago Domen made Cachix work only with recent Nix (`>2.3`), and Nix installer installed two years old (`2.0.3`), so or Cachix goes because it is rightfully asking for fresh Nix release and CI builds are without the project caches, or update the Nix right after installing Nix. That is why it updates in the configuration.

I was updating it with `nix update-nix` but now CI broke, and old way is probably stil works more properly. I got tired on Nix tooling, there should be only one way to update the Nix, and the installer should just work, and always install the latest release, so it does not need updating right after installing, and it should accept a trusted users list, so people not need to debug it and figure-out that the service restart is neended right after the installation.

Anton-Latukha:
I am so tired of fixing the installation issues of Nix (┛ಠ_ಠ)┛彡┻━┻

It is a package manager (installer) that can not install itself on so many systems or operate on them normally after the installation for the last number of years.

Today is just a day when I feel more tired and care less about debugging the Nix installer.

The macOS migration hack after Nix update now is not needed, but the service restart is still needed.

Anton-Latukha:
Also isn't that logical that CI testing system always should load the latest stable release of Nix, as we install, use and test with current package versions, not legacy ones.

I'd preferred if there always loaded the latest container, but that is impossible in Travis, as I know there containers also need to go through presetup.

Anton-Latukha:
It is great that all this happened before I published an article and repo for Haskellers community how to do CI with Nixpkgs, now I do not need to explain installer issues and I can pretend that Nix would work further.

Anton-Latukha:
Wow.

Linux currently works only old `nix-channel --update && nix-env -iA nixpkgs.nix`.
On macOS currently works only `sudo nix upgrade-nix`. While normal user is trusted and service was restarted.

So there is this, so I can not pretend it works normally.

PR #639, opened 2020-06-20
CI: .travis.yml: clean-up

Anton-Latukha:
Just some basic clean-up.

Strict build logically should be tested once, and on the most modern GHC possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,10 +91,9 @@ jobs:
         - linkWithGold=true
       os: linux
       dist: bionic
-    - name: GHC 8.6.5, Linux, Strict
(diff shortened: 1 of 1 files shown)


sjakobi (approved):
Cheers!

Anton-Latukha:
Cheers, @sjakobi.

This is still a very a basic work. I am just happy and getting ready to form a useful repo from these files to boost the Haskell public.

I need to fix my Haskell toolchain setup, dodge `nixos-unstable` upstream, IDK.

My hope is then when all these ends and the dust settles - Functional configuration and deployment would be better and more reliable, also by having actual development community, and that Nixpkgs effort would be reused.

PR #643, opened 2020-06-23
default.nix: filter package root with nix-gitignore

sorki:
so we don't copy stuff like `dist-newstyle`.


Anton-Latukha:
Hey, yes, nice point.

Issue #645, opened 2020-06-23
Please, deduplicate the docs in `.travis.yml`, `build.sh`, `default.nix`


Anton-Latukha:
I worked with files in the above order. And it seemed logical to provide the docs on every stage.

I agree with guys, restating docs three times is too verbose. Especially when it is done and `.travis.yml` relies on `build.sh` interface, which relies on `default.nix` interface, which relies on Haskel Lib.

I do not think slashing the docs completely is sane, since people who do not know Nixpkgs should be guided simply and intuitively. And we can not point to the Haskel Lib directly, the functionality of it is simplified in `default.nix`, many wrappers of Lib turned into simple boolean switches in `default.nix` header, and Nixpkgs docs there are bigger or sometimes not existent, more assuming and sometimes more obscure, then simply stating/explaining. So I think living the explanations in `default.nix` and make pointers to it is a sufficiently elegant solution.

Anton-Latukha:
All this work does not go in vain, since during all this rehashing and rewriting the documentation and code always simplifies and improves a bit, which is important point we aim for to release this to make #637.

Issue #637, opened 2020-06-19
Please, allow to form CI results into separate repository


Anton-Latukha:
The results are portable and should be shared to Haskell community.
This `default.nix` or making it an `*.nix` extension to package allows exposing all `haskell.lib` functionality.
It is very practical useful thing and nobody shared it before with a community - it should be shared.

1. https://github.com/haskell-nix/hnix/blob/master/default.nix which exposes `Nixpkgs haskell.lib` API - can be used by itself for development (it also integrates with Haskell IDE Engine), debugging, build testing, package quality assurance.
2. https://github.com/haskell-nix/hnix/blob/master/build.sh which exposes `Nixpkgs haskell.lib` API - can be used with `default.nix` to have also GHCJS build.
3. https://github.com/haskell-nix/hnix/blob/master/.travis.yml which exposes `Nixpkgs haskell.lib` API.

I would get the git history and form the separate repository, that would also allow to clean the configuration and make it universal abstraction for other projects to use.

I would do this in the nearest time.

I think it is logical to host Haskell Nix CI setup result in `haskell-nix` group.

Anton-Latukha:
Kindly asking for a response and informing, since that code is not only mine code and work.

I hope #636 would be accepted. Being socially responsible requires me to publish the work, so the Haskell community would be able to easily build/extend their CIs leveraging the Nixpkgs Haskell Lib features have ability to do testing for Nixpkgs in any CI even if they do not know any Nix, and those people are the major part of Haskell community.

Anton-Latukha:
Due to no permissions to form related repository here, I need to create repository `haskell-nixpkgs-ci` separately.

Notifying @jwiegley @mightybyte  @domenkozar, you are the main base authors on top of code of which I did work and added ideas on top in relevant files, I would respect the BSD 3 and preserve commit history and work of authors.

Anton-Latukha:
The repository is at: https://github.com/Anton-Latukha/haskell-with-nixpkgs

jwiegley:
@Anton-Latukha Are you asking for a review of the code in that repository?

Anton-Latukha:
@jwiegley I know that BSD license does not require it, but as a human simply and politely notifying that the part of the work is formed into a separate repository, and good if it was possible to have it in the `haskell-nix` group. That is all to it.

[...]

PR #647, opened 2020-06-23
CI: refactor CI setup config

### LINKED ISSUE #648: Please, wait, debug and make newly released Nix installer work for the CI setup
Well, I was looking into a crystal ball in https://github.com/haskell-nix/hnix/issues/641#issuecomment-648333812.

I was doing build tests in #647, and setup just broke because of the rollout of the Nix installer release.

New Nix installer release was awaited for a couple of years, from when Nix 2.0 was released there were no installer releases. macOS problems and Cachix version demands forced quick to update the installer release.
While they rolled-out and would debug it for some time - it broke our CI for the nearest time.

I would figure-out what is changed, and how to adjust to the new Nix installer.

Anton-Latukha:
Working gradually, need to test CI passage.

Now when most docs are removed, it is more obvious where to reduce the NOTEs.

Also synchronized the interface explicitness with what `build.sh` provides.

Would test and maybe add a new build feature check.

Working towards  #645 and #637.

Renamed `(GHCVERSION->compiler)` to correspond option with the `nix-build` and Nixpkgs Haskell Lib.

---

In the middle of work Nix installer got a major release, so also includded a bigger CI refactoring to keep CI working #648


Anton-Latukha:
Well, I was looking into a crystal ball in https://github.com/haskell-nix/hnix/issues/641#issuecomment-648053261

Nix upstream after several years of waiting just made and new release or Nix installer, at last. And it seems that it uses macOS originated installer code.

Well, while they would debug the new Nix installer, it broke our CI for the nearest time.

Anton-Latukha:
What happened, basically Nix upstream started to use multiuser installer by default, and that broken sandboxing in the config, and since unpure env went into Nix inputs - which triggered the whole stack rebuild.

Anton-Latukha:
Holy Moly, it works again.

The migration process is too rough, there are hacks removed, and some hacks added. It is impossible to gradually commit them, since they can not be split into stages, since all them are pretty much a single atomic change to make a proper working new Nix from new install.

PR #649, opened 2020-06-23
CI: .travis: rm sudo, enable container use

Anton-Latukha:
Lets look at timings, maybe it would be faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ version: ~> 1.0
 
 language: nix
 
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
Not enough time reduction to compensate for the drop in RAM for Haskell GHC builds:
`master`: https://travis-ci.org/github/haskell-nix/hnix/builds/701456933
`PR`: https://travis-ci.org/github/haskell-nix/hnix/builds/701461488

PR #660, opened 2020-06-26
Add cabal.project, only .gitignore cabal.project.local

sorki:
This prevents accidental surprises like the one I've managed to hit
today - `git worktree add master` and cabal build in master directory
tries to look-up `cabal.project` but only discovers one in parent
directory and starts reusing `dist-newstyle` for the build.

Better to have this explicit which devs can also use as a template to
add other temporary deps.

Edit: Reflect edited commit message typos.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@
 /nix-test-eval*
 /nix/
 TAGS
-cabal.project*
+cabal.project.local
 ctags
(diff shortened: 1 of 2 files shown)


sjakobi (approved):
`git worktree`! More evidence you're a wizard… :raised_eyebrow:

PR #678, opened 2020-07-15
cabal: clean-up of dep bounds

Anton-Latukha:
Clean-up.

### DIFF
--- a/hnix.cabal
+++ b/hnix.cabal
@@ -993,18 +993,18 @@ test-suite hnix-tests
   build-depends:
       Diff
     , Glob
-    , base >=4.11 && <5
+    , base
     , base16-bytestring
     , bytestring
     , containers
     , data-fix
-    , deepseq >=1.4.3 && <1.5
+    , deepseq
     , directory
     , exceptions
     , filepath
     , hedgehog
     , hnix
-    , megaparsec >=7.0 && <8.1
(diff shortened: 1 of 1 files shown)


hnix.cabal
```diff
@@ -940,31 +940,31 @@ executable hnix
...
-    , bytestring
-    , comonad
-    , containers
-    , data-fix
-    , deepseq
-    , exceptions
-    , filepath
-    , free
+      hnix
+    , aeson >= 1.4.2 && < 1.6
```
mightybyte: You don't have to duplicate bounds that already exist elsewhere in the cabal file.  You only need to put bounds for the dependencies here that don't have bounds elsewhere.

mightybyte (changes requested):

hnix.cabal
Anton-Latukha: Thanks for curation.

Did `cabal gen-bounds` for executable and then added `library` override.

Would redo.

Well, then would also check/remove other bound copies.

Anton-Latukha: Thanks, indeed - bounds were normal, additionally made bounds unique.

mightybyte (approved):

sjakobi (approved):
Thanks! :)

PR #685, opened 2020-08-03
Update to badges

Anton-Latukha:
Old CI badge no longer reported right status of things, so updated our badge monitoring.

CI badges can be only per workflow, well actually that is pretty informative concise status reporting.

Organized badges in the table.

Result of the PR can be seen [here](https://github.com/Anton-Latukha/hnix/tree/2020-08-03-badges).


sjakobi (approved):
That's a lot of badges! ;)

Cheers!

Anton-Latukha:
Since we have badges, we may remove one of two: CI notifications to Email or badges.

I personally up to remove Email notification and have a badges, sine nevertheless it is not so much critical to notify on Email the minute status changed, and I during work open HNix front page from time to time to open subsections and automatically glance over the colors of badges.

And since we monitor `Release dependencies` with RSS, and status is badge is always `outdated`, and upstream does not fixes the respect to versions, we may cut the badge complettely, and rely on us reading RSS for the time being.

The badges are better in that why that they are on demand, for example, check and open the dependencies report when you in fact interested in it before release.

PR #694, opened 2020-08-14
Merge of haskell-with-nixpkgs progress into hnix

Anton-Latukha:
Providing "default" value to `{compiler,rev}` which tracks default Nixpkgs GHC
and local channel, accordingly.

Clean-up.

Properly providing secret to the GitHub CI workflows.

Project can free form Travis CI.

rm of the old hacky GHCJS `build.sh` for Travis CI, build can be replaced by
`ghcjs-build.sh` from `haskell-with-nixpkgs` on wish.

In terms of the fish shell the rebase merge is done in this way:

```fish
set -l tmpDir "$HOME/src/tmp"
set -l sourceDir 'haskell-with-nixpkgs'
set -l sourceUrl "git@github.com:haskell-nix/$sourceDir.git"
set -l masterDir 'hnix'
set -l masterUrl "git@github.com:haskell-nix/$masterDir.git"
set -l fileList \
default.nix \
build.sh \
ghcjs-build.sh \
Cabal-Linux.yml \
Nixpkgs-GHCJS.yml \
Nixpkgs-Linux-additional.yml \
Nixpkgs-Linux-maim.yml \
Nixpkgs-macOS.yml
mkdir -p "$tmpDir"
rm -rf "$tmpDir/$sourceDir"
rm -rf "$tmpDir/$masterDir"
cd "$tmpDir"
git clone "$sourceUrl" "$sourceDir"
cd "$sourceDir"
git-pass-filter $fileList
cd "$tmpDir"
git clone "$masterUrl" "$masterDir"
cd "$masterDir"
git checkout -b upd-from-"$sourceDir"
git remote add hn "$tmpDir/$sourceDir"
git fetch hn master
git merge -S --allow-unrelated-histories hn/master
```

7 files changed, 385 insertions(+), 103 deletions(-)
.github/workflows/Nixpkgs-GHCJS.yml            |  33 +++++
.github/workflows/Nixpkgs-Linux-additional.yml |   5 +-
.github/workflows/Nixpkgs-macOS.yml            |   6 +-
build.sh                                       |  98 ++------------
default.nix                                    |  30 +++--
ghcjs-build.sh                                 | 143 ++++++++++++++++++++
silenced-lazy-ghcjs-build.sh                   | 173 +++++++++++++++++++++++++

PR #695, opened 2020-08-15
updating to aeson, negotiating [From,To]JSON instances

Anton-Latukha:
I tried to use new `aeson` in `default.nix` - that required ~5 more overrides and resulted in infinite recursion, so I decided to stick to main agenda - follow Hackage and sync with `aeson`, and approach Nixpkgs infrastructure after.

### DIFF
--- a/hnix.cabal
+++ b/hnix.cabal
@@ -384,7 +384,7 @@ library
       src
   ghc-options: -Wall -fprint-potential-instances
   build-depends:
-      aeson >= 1.4.2 && < 1.6
+      aeson >= 1.4.2 && < 1.5.3
     , array >= 0.4 && < 0.6
     , base >= 4.11 && < 5
     , binary >= 0.8.5 && < 0.9


sjakobi (approved):
:+1: for the library change. Cheers!

Anton-Latukha:
Chicken and egg problem.

I don't want to merge #694, since it is not ready and since I prioritize #695, which builds too slow since Cabal Hackage builds in CI run every time fully anew. And cache in #696 does not save since build does not succeed, because of #695.

So, I need to bite the bullet and pull/build everything with Cabal locally.

Anton-Latukha:
So far with persuing using `aeson` available instances, and enabling back `ToJSON1` HNix instances:
```
src/Nix/Expr/Types.hs:1:1: error:
    Exception when trying to run compile-time code:
      Constructor ‘DynamicKey‘ must only use its last 1 type variable(s) within the last 1 argument(s) of a data type
CallStack (from HasCallStack):
  error, called at ./Data/Aeson/TH.hs:1895:35 in aeson-1.4.7.1-PoZwsvVZAhEzlieMP6aSv:Data.Aeson.TH
    Code: deriveJSON1 defaultOptions ''NKeyName
  |
1 | {-# LANGUAGE BangPatterns #-}
```

When:
```haskell
-- | A 'KeyName' is something that can appear on the left side of an
-- equals sign. For example, @a@ is a 'KeyName' in @{ a = 3; }@, @let a = 3;
-- in ...@, @{}.a@ or @{} ? a@.
--
-- Nix supports both static keynames (just an identifier) and dynamic
-- identifiers. Dynamic identifiers can be either a string (e.g.:
-- @{ "a" = 3; }@) or an antiquotation (e.g.: @let a = "example";
-- in { ${a} = 3; }.example@).
--
-- Note: There are some places where a dynamic keyname is not allowed.
-- In particular, those include:
--
--   * The RHS of a @binding@ inside @let@: @let ${"a"} = 3; in ...@
--     produces a syntax error.
--   * The attribute names of an 'inherit': @inherit ${"a"};@ is forbidden.
--
-- Note: In Nix, a simple string without antiquotes such as @"foo"@ is
-- allowed even if the context requires a static keyname, but the
-- parser still considers it a 'DynamicKey' for simplicity.
data NKeyName r
  = DynamicKey !(Antiquoted (NString r) r)
  -- ^
  -- > DynamicKey (Plain (DoubleQuoted [Plain "x"]))     ~  "x"
  -- > DynamicKey (Antiquoted x)                         ~  ${x}
  -- > DynamicKey (Plain (DoubleQuoted [Antiquoted x]))  ~  "${x}"
  | StaticKey !VarName
  -- ^
  -- > StaticKey "x"                                     ~  x
  deriving (Eq, Ord, Generic, Typeable, Data, Show, Read, NFData, Hashable)
```

I am new to strict evaluation in Haskell, continuation...

Anton-Latukha:
Most probably I need to try to not derive it, learn how with `aeson` to build instance manually to understand something.

Anton-Latukha:
Again found the good old: https://github.com/haskell/cabal/issues/6228#issuecomment-553264551 on NixOS.

Again and again over the years.

"This is a tricky situation..."

Anton-Latukha:
I arrived at the same situation as `sjakobi`.

Wrapping this PR up to partial fix, as `sjakobi` - would restrict `aeson` version for now, this would resolve the https://github.com/haskell-nix/hnix/pull/695#issuecomment-674410671 and release situation.
When someone gets a better understanding - then it can be solved properly.

PR #712, opened 2020-09-12
Options/Parser: add '--long-version' that adds debug information

Anton-Latukha:
Option for debugging and bugreporting purposes.

Current output:
```
Version: 0.9.1
Commit: 2dc211314eee32e02f83de16062a5c1355e11fc7
  date: Sat Sep 12 13:31:59 2020 +0300
  branch: 2020-09-12-add-GitRev-to-version
```

Also date and branch is to direct contributors attention to updating (or having
in mind) that.

M  hnix.cabal
M  src/Nix/Options/Parser.hs


Anton-Latukha:
Branch information probably is excessive, it is probably should be reduced by YAGNI principle. But well, if project would diverge into several branches (which is not likely). So the branch report is YAGNI.

Anton-Latukha:
There is a way to extract the version informaiton into into the executable using Template Haskell, that should untie HNix from Cabal machinery in this particular place.

But there seems to be no package/module to do that the nice way.

I sorta do not want to drag [`unbeliever`](https://github.com/aesiniath/unbeliever) [`core-program`](https://hackage.haskell.org/package/core-program) into HNix just for [`Core.Program.Metadata`](https://hackage.haskell.org/package/core-program-0.2.4.5/docs/Core-Program-Metadata.html) to retrieve 'Verison` in a somewhat more portable form, so for now it left in a more simple Cabal metadata manner but with comment that the Template Haskell is possible and in the far future preferred there.

Anton-Latukha:
This PR seems straight-forward.

I would remove `branch` report from it, and include the rest into today's `0.11.0`.

Anton-Latukha:
Well, `nix-build`ing gives:

```
> hnix --long-version

Version: 0.10.0
Commit: UNKNOWN
  date: UNKNOWN
```

So the Template Haskell returns info only for the canonical `cabal` builds.

`nix-build` AFAIR strips the source code before including it into closure compilation processes, which means all SVC/Git metadata stripped from it also, so there is nowhere to get it from during the `nix-build`. The bummer... But there is no way to avoid this optimally easily.

But the ideal should not be the enemy of the good. If people use `nix-build`, we can just ask them to give us the `cabal` from `nix-shell`, which is a sort of close approximation, and if fields are empty, well - they are just additional maybe helpful info that happened to be empty. We just can request `nix-shell` `cabal` build info, which roughly approximate the case. We also may request people to try to provide commit hash manually, and UNKNOWN sort-of encourages it, if they can do it.

The most important that canonical Cabal works.

PR #717, opened 2020-09-13
Options/Parser: debugVersionOpt: comment: add `nix-build` note

Anton-Latukha:
[Remark](https://github.com/haskell-nix/hnix/pull/712#issuecomment-691654639) deserves a comment note in the source code.

### DIFF
--- a/src/Nix/Options/Parser.hs
+++ b/src/Nix/Options/Parser.hs
@@ -199,6 +199,7 @@ versionOpt = shortVersionOpt <*> debugVersionOpt
     <> help "Show release version"
     )
 
+  --  2020-09-13: NOTE: Does not work for direct `nix-build`s, works for `nix-shell` `cabal` builds.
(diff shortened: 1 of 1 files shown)


alexfmpe:
Ran into this exact issue on a `--version` for a CLI.
One can workaround by wrapping the executable and passing git data as environment variables:

```nix
root = ./.;

exe = nixpkgs.writeScript "my-package-exe" ''
  #!/usr/bin/env bash
  set -euo pipefail
  export REV=$(git -C ${root} rev-parse HEAD)
  export BRANCH=$(git -C ${root} rev-parse --abbrev-ref HEAD)
  ${nixpkgs.haskell.lib.justStaticExecutables pkgs.my-package}/bin/my-package $@
';
```

alexfmpe:
Downside is that it breaks `--version` for `nix-shell` `cabal`, though maybe more wrapping could fix that

PR #719, opened 2020-09-13
cabal, CHANGELOG: 0.10.1 release, unifying the changelog

Anton-Latukha:
* "Misc" use was disputed, moving back to use of the "Additional" category.

* Using a (link) pattern for URLs, reasons being:
  * Including the markup into the outlines or hyperlinks considered a
    bad practice and should be used carefully.
  * The human cause:
    The reason for the changelog is so people read through it and get the info.
    Hyperlinks make text colored. Studies of readability have shown that
    people perceive the information of a brightly colored text
   much more poorly than the text of regular coloring.
    And delivering the meaning is what changelog text is for, so the
    informative text should be regularly colored.
  * The technical cause -
    the Hackage does not parse the markup inside the hyperlinks:
    https://hackage.haskell.org/package/hnix-0.10.0/changelog
    Probably because of the reason *1, but Hackage reason alone is enough.
  * Since people perceive and think about the information they are currently reading
    mostly linearly (so in the FIFO manner), so if `(link)` is put before information text,
    they are are placed in the same uniformal space and look more uniform upon reading.
    And upon reading the "link" context gets pushed first from the FIFO mind buffer and
    the further understanding the text the person is not distracted by
    sudden reading (taking-in) of the colored word "link" in the end of the
    semantically challenging text that is required to be understood.

* Dependency requirements of last major versions made literal.

* Link to `(diff)` made more obvious. While professionals tend to look for the
link to the full diff, diff under version number is ambiguous to where the
version number link leads, it can be for example: a GitHub release (where
different forms of packages are provided), diff, commit listing (as it was before
the current change), official forum post, official news... etc.

* Made `(diff)` link to point directly towards the diff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,44 +1,62 @@
+
 # Changelog
 
-## [0.10.0](https://github.com/haskell-nix/hnix/compare/0.9.1...0.10.0) (2020-09-12)
+### [(diff)](https://github.com/haskell-nix/hnix/compare/0.10.0...0.10.1#files_bucket) 0.10.1 (2020-09-13)
+
+* Additional:
+  * [(link)](https://github.com/haskell-nix/hnix/pull/715/files) `{Binding, NExpr, NExprF, NKeyName}` gained `Ord1` instances.
+    * These instances were required by downstream projects to be able to use newer HNix.
+  * [(link)](https://github.com/haskell-nix/hnix/pull/712) CLI gained `--long-version` option for gathering a detailed debug information.
+    * Currently, reports Git commit and its date.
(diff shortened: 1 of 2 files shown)


Anton-Latukha:
[(link)](https://github.com/haskell-nix/hnix/blob/2020-09-13-form-0.10.1/CHANGELOG.md) - to observe the result of the changes.

Anton-Latukha:
Now I consider this ready.

How does it look?

Since this is an update fix release - would follow with the release.

Be free to open and merge your and others PRs with changes to the changelog or docs.

PR #730, opened 2020-09-29
default.nix: switch to the GHC 8.10.1, now executable gets build

Anton-Latukha:
Switching Nix dev env and its CI to GHC 8.10.

To avoid contribution loop problems, since we pin Nixpkgs in default.nix - it is logical by default CI to follow the supported way of building.

The downside is - the maintainer would need to solve Nix issues when wants to update the Nixpkgs revision, or play a wait and pick game with Nixpkgs.

The other downside observed - on rev update, the `Nix-shell & Hoogle` CI build needs to rebuild the whole Haskell stack from the bottom up, so the rev update process takes a lot of wait time (hours), and also should be done from repo internal branch to save the already built parts of the stack during rev testing.

### DIFF
--- a/.github/workflows/Optional-Nix-dev-env-macOS.yml
+++ b/.github/workflows/Optional-Nix-dev-env-macOS.yml
@@ -13,7 +13,7 @@ on:
 
 
 env:
-  rev: "nixos-unstable"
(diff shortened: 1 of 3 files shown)


Anton-Latukha:
To build-up cache reopening PR through an internal branch.

Anton-Latukha:
#731

PR #741, opened 2020-10-23
Fix QQ Text lifting error

expipiplus1:
Work around https://gitlab.haskell.org/ghc/ghc/-/issues/12596 using
https://stackoverflow.com/questions/38143464/cant-find-inerface-file-declaration-for-variable


expipiplus1:
This has been working very well for me.

PR #744, opened 2020-10-25
Use pre-whitespace position for source end locations

### LINKED ISSUE #743: Location annotation for expressions includes comments
```
{ } # hello
^---------^ is the annotated location
^-^ is what I would expect the annotation to be
```

This is primarily because most parsers use `annotateLocation1` outside of `lexeme p = p <* whitespace`

Some solutions:

- Float `lexeme` out beyone `annotateLocation1`
- Maintain some state in the parser regarding the latest non-whitespace position

expipiplus1:
Fixes https://github.com/haskell-nix/hnix/issues/743

~I have not tested this extensively, so please consider this a WIP~
I have tested this a little bit and it seems to work

### DIFF
--- a/src/Nix/Parser.hs
+++ b/src/Nix/Parser.hs
@@ -53,13 +53,13 @@ import           Control.Applicative     hiding ( many
 import           Control.DeepSeq
 import           Control.Monad
 import           Control.Monad.Combinators.Expr
+import           Control.Monad.State.Strict
 import           Data.Char                      ( isAlpha
                                                 , isDigit
                                                 , isSpace
                                                 )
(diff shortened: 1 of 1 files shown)


layus (approved):
Looks okay to me. Nice work :-).

(But I do not have commit access sadly)

expipiplus1:
Lol, can't believe it took me 4 years to come up with this. https://github.com/haskell-nix/hnix/pull/35#issuecomment-228556393

(I think back then it might have been hampered by using trifecta instead of megaparsec or something)

expipiplus1:
Thanks @layus

Issue #747, opened 2020-11-01
`shell.nix` is currently vacuous


Anton-Latukha:
It is well known, `...` does not work as intended in Nix, since implicit variadic attributes can not be passed during imports of other files (it is considered a feature of the language).

For example, both versions:

```
attrs@{ ... }: (import ./. attrs).env
# is equivalent to
attrs@{} : (import ./. attrs).env
```

, they equivalent since they both behave the same, they d not pass any `attrs` into the derivation:

```fish
nix-shell --argstr compiler 'ghc884'    # Does not work in both above versions of `shell.nix`
ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.10.2
```

While `nix-build` that loads the `default.nix` - works perfectly:

```fish
nix-build --argstr compiler 'ghc884'    # Works properly, because uses `default.nix` interface
```

So I question the binding and passing of `attrs` in `shell.nix` at all - because it does not pass any implicit args in any case of `{}` or `{...}` or `{ <explicit_args>, ...}` - it is a rope of sand that trips people.

Since `attrs` does not work - the `shell.nix` is fully vacuous and by the Occam's razor can be removed, and so simply `default.nix` is used, load env from there - then `default.nix` API becomes accessible for `nix-shell`. But that API is also for compilation and many keys do not apply for the shell env.

The least what can be done - is provide explicit interface in `shell.nix` - there is not much attrs for it currently, it is two attrs.

layus:
Closing since it seems fixed (shell.nix provides defaults for the shell, and access to the full API).
Feel free to reopen if need be.

sorki:
The arguments passing bug that reappears from time to time is so annoying we should devise a test for this at some point for `hnix`.

```
nix-build --argstr compiler 'ghc884'    # Works properly, because uses `default.nix` interface
nix-shell --argstr compiler 'ghc884'    # Does not work in both above versions of `shell.nix`
```

I've been digging into this for couple hours (there are numerous PRs claiming to fix this) and I think @Kiwi knew of a fix and one required addition to `shell.nix` for recent Nix but is too shy to submit a PR.

Anton-Latukha:
:arrow_up:
Indeed, this report is opened since the code does seem like it passes the interface, but the interface passing to `shell.nix` does not work.

[...]

PR #748, opened 2020-11-01
shell.nix: use Haskell notation, explicit attrs

Anton-Latukha:
This form is 1:1 Haskell.

#747 explains the addition of explicit attr into `shell.nix`.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -76,8 +76,8 @@
 , executableNamesToShellComplete ? [ "hnix" ]
 
 
-# Include Hoogle into derivation
-, withHoogle  ? true
+# Include Hoogle executable and DB into derivation
+, withHoogle  ? false
 
 
 # Nix by default updates and uses locally configured nixpkgs-unstable channel
--- a/shell.nix
+++ b/shell.nix
@@ -1,1 +1,5 @@
-{} @ attrs: (import ./. attrs).env
+attrs@
+  { compiler ? "ghc8101"
+  , withHoogle ? true
+  }:
+(import ./. attrs).env
--- a/shell.nix
+++ b/shell.nix
@@ -1,5 +1,6 @@
 attrs@{...}:
 let defaultAttrs = {
+  # Defaults are put in this form deliberately. Details: #748
   withHoogle = true;
   compiler = "ghc8101";
 };
--- a/shell.nix
+++ b/shell.nix
@@ -1,5 +1,6 @@
-attrs@
-  { compiler ? "ghc8101"
-  , withHoogle ? true
-  }:
-(import ./. attrs).env
+attrs@{...}:
+let defaultAttrs = {
+  withHoogle = true;
+  compiler = "ghc8101";
+};
+in (import ./. (defaultAttrs // attrs)).env


Anton-Latukha:
Well.

Great.

`nix-shell --pure --command 'hoogle True'`

`shell.nix`:
```nix
attrs@
  { compiler ? "ghc8101"
  , withHoogle ? true
  }:
(import ./. attrs).env
```

`default.nix`
```nix
{ withHoogle ? false }
```

Result: `hoogle: command not found`, since: `withHoogle false`...

How.
Why?
How it happens that overshadowed expression gets evaluated instead of what shadowed it.

I'm sorry, I thought lazy evaluation should start from evaluating what shadows the expressions/values. Solving the expression starts with `call-by-name` expression getting fully composed, and then called-by-name `withHoogle` value gets solved. At that point, import already happened without any value evaluation in the expression composition phase, so `default.nix` already got loaded into `shell.nix` scope and the `default.nix` `withHoogle` expression got overshadowed by the binding over it the `shell.nix: withHoogle ? true` expression, which all happens before any `withHoogle` value resolution starts, or I do not understand how laziness works properly.

This situation together with `...` not passing interface - I do not see how one can reasonably build & pass any reasonable API in Nix to the user.

Anton-Latukha:
Am I missing something obvious?

layus:
What you are missing is not obvious at all. Debugging with builtins.trace helped me to understand what happens.
`attrs@` capture the actual attset given as argument. Default parameter values like `attrs@{withHoogle ? true}` are ignored.

```nix
let 
  funA = attrs@{withHoogle ? true} : {inherit attrs withHoogle;};
  funB = attrs@{...}: let attrsWithDefault = { withHoogle = true; } // attrs;
                      in { attrs = attrsWithDefault; withHoogle = attrsWithDefault.withHoogle; }; 
in {
  a = funA {}; 
  b = funB {};
}
```

```json
{
  "a": {
    "attrs": {},
    "withHoogle": true
  },
  "b": {
    "attrs": {
      "withHoogle": true
    },
    "withHoogle": true
  }
}
```

I took the liberty to add one commit to this PR. Feel free to edit/delete/rebase it at will.

Anton-Latukha:
Thank you for confirming and adding a workaround.

I am sure I knew it already and encountered it a couple of times before. But active search not successful in confirmation.

& without other people confirming such things, strange quirks in the techs/languages feel like an own lack of something or hallucinations.

Since the project needs to have B2B (I mean, bug-to-bug) implementation, I would do some for streamlining this agenda socially, "must ... collect ... all ... bugs" into the proper article in the proper Wiki. Also HNix probably needs to abstract quirks/bugs so those places explicitly use additional special module abstraction, or at least denoted in the implementation. So the project would be able in the future to keep compatibility and be able to have language dialect refactoring.

layus:
I was actually quite surprised to discover that behavior. It make some sense, but is rather counter-intuitive.

layus:
Oh, there is a big orange box in the doc about that. We are no the first ones to stumble upon this :-)
https://nixos.org/manual/nix/stable/#ss-functions

Anton-Latukha:
This use, or as it is put in manual:
```
let
  function = args@{ a ? 23, ... }: args;
in
 function {}
# will evaluate to an empty attribute set. 
```
- such a case should notify with the message, warning, or the full solution is simple - use of `?` in `@` - should be deprecated, because as it officially put - does not work.

Anton-Latukha:
The reason why one needs to put:
```nix
attrs@{compiler ? "ghc8101", ...}:
```

While as we saw - it does not work, and manual puts it is vacuous, nevertheless, suddenly it provides the most useful side effect - it the only case I know when command argument gets respected and passed into derivation:
```
nix-shell --pure --argstr compiler "ghc884" --command 'ghc --version'
```
Would do what user asked - `8.8.4`.

How we currently done:
```nix
attrs@{ ...}:
let defaultAttrs = {
  # Defaults are put in this form deliberately. Details: #748
  withHoogle = true;
  compiler = "ghc8101";
};
in (import ./. (defaultAttrs // attrs)).env
```

Surprises with old behavior:
```
nix-shell --pure --argstr compiler "ghc884" --command 'ghc --version'
```
Would return *not* asked `8.8.4`, but "default" `8.10.1`.
Old: `attrs@{ ... }:  (import ./. attrs).env` also not passed the arg, what tried to avoid in the first place (the main initial reason for the PR).

So - it seems like it needs a default set in both places to work as expected.

```
attrs@{compiler ? "ghc8101", ...}:
let defaultAttrs = {
  # Defaults are put in this form deliberately. Details: #748
  withHoogle = true;
  compiler = "ghc8101";
};
in (import ./. (defaultAttrs // attrs)).env
```

Works as expected:
```
nix-shell --pure --argstr compiler "ghc884" --command 'ghc --version'
```
Would supply asked `8.8.4`.
And default case:
```
nix-shell --pure --command 'ghc --version'
```
Would supply the default case, `8.10.1`

Very strange, this needs to be documented better.

PR #749, opened 2020-11-01
CHANGELOG.md: add 0.11.0

Anton-Latukha:
@expipiplus1 

If it looks horrible - it is because it is being a Helloween present.

You are free to proactively modify the PR in any form you like. Edit it directly in this PR, add fixes on top or force push your commit, or copy it, or open your version.

The main thing in the text I attached to is the form of it, and that can be reasoned about also.

The result is here: https://github.com/haskell-nix/hnix/blob/2020-11-01-preparing-v0.11.0-release/CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,22 @@
 
 # Changelog
 
+
+## [(diff)](https://github.com/haskell-nix/hnix/compare/0.10.1...0.11.0#files_bucket) 0.11.0 (2020-11-02)
+
+* Breaking:
+  * [(link)](https://github.com/haskell-nix/hnix/pull/740) Deleted incorrect `instance Generic1 NKeyName` from `module Nix.Expr.Types`.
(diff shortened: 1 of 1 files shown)


expipiplus1:
Very detailed! LGTM

expipiplus1:
I would suggest adding a '## WIP' section to the changelog to track changes as they happen, to prevent having to remember everything that happened between versions.

Anton-Latukha:
True.

a) Making people write into a changelog is the best information pass through, but it is additional enforcing discipline on volunteers. And in the end, something would sleep through the cracks and maintainers would need to form a changelog.

b) In the way how readme TOC currently autogenerates. It is possible to populate WIP changelog using merged commit names and docs, and CI also can collectet the links to the PRs. That also would require before release that automatic changelog manually gets cleaned and refactored - but that is a way easier, when it is already populated with relevant info. And that autogeneration would require/encourage people to put info into commits and request that from them more - that is a better practice since it is not mandatory while encourages everyone, when explanations of PR name and commit name is not enough, to put explanations into commit message, which also would give way better history/blame/debug exp. This is a long-term idea - automation of changelog is more complex than TOC generation, but there is some relevant CI actions already provided and I have a part of the solution (merge abstractions) in the implementation of two-pass git bisect.

PR #756, opened 2020-11-16
Make `->` right assosciative

layus:
As per the nix source code https://github.com/haskell-nix/nix/blob/61e816217bfdfffd39c130c7cd24f07e640098fc/src/libexpr/parser.y#L296-L307

The bug was found while parsing firefox https://github.com/NixOS/nixpkgs/blob/5030e5cdc7f4782e895c470e8ea9fa574274b223/pkgs/applications/networking/browsers/firefox/common.nix#L76-L77


expipiplus1:
Thanks!

PR #757, opened 2020-11-16
Various fixes and bugfixes encountered while testing

layus:


layus:
This should be fairly friction free.

src/Nix/Builtins.hs
```diff
@@ -244,6 +246,16 @@ builtinsList = sequence
...
+  {-
+  , add  Normal   "fetchGit"         fetchGit
+  , add  Normal   "fetchMercurial"   fetchMercurial
+  , add  Normal   "filterSource"     filterSource
+  , add  Normal   "fromTOML"         fromTOML
+  -}
+  , add  Normal   "getContext"       getContext
+  {-
+  , add  Normal   "path"             path
+  -}
```
expipiplus1: Is this list supposed to be alphabetized?

the TODO elements and 'nixPath' are out of order.

src/Nix/String.hs
```diff
@@ -231,6 +233,16 @@ runWithStringContextT :: Monad m => WithStringContextT m Text -> m NixString
 runWithStringContextT (WithStringContextT m) =
   uncurry NixString <$> runWriterT m
 
+-- | Run an action that manipulates nix strings, and collect the contexts encountered.
+-- Warning: this may be unsafe, depending on how you handle the resulting context list.
```
expipiplus1: Is it possible to elaborate a little on what unsafe means, and what handling would cause bad behavior?

expipiplus1 (dismissed):
Looks good, a couple of minor nits, but otherwise I'd be very happy to merge, thanks!

src/Nix/Builtins.hs
layus: Oh that's funny. Something tampered with my internal Ord String instance :-)

src/Nix/String.hs
layus: Well, as soon as you separate the context from the string, you are responsible of doing something "principled" with both the context and the string.

runWithStringContext ensures that the monadic context is merged in a nix string in the end. So you are sure that the contexts you abstracted away end up somewhere sensible. Look at "hnix/src/Nix/{Json,XML}.hs". They take a nix value, extract the strings with `extractNixString` and the WithStringContext ensures that the resulting string gets wrapped in all the contexts that have been encountered during the computation.

This version however makes it possible to just discard these contexts from the resulting tuple.

Any idea how to better phrase the comment ?

src/Nix/Builtins.hs
layus: Turns out I was not completely blind. The order was not correct anyway. Fixed in https://github.com/haskell-nix/hnix/pull/757/commits/1a3a2abbc4e091d3b666578ed313488aa8e10641

src/Nix/String.hs
expipiplus1: That sounds good to me, I guess these are going to be be quite infrequently used by an end user anyway!

layus: I hope so :-).

In fact this one should probably move to the #554 PR, which will probably be the sole usage for it.
{Json,XML}.hs are the sole usages for WithStringContext. Not really a common tool.

expipiplus1: Sure, no bother either way :)

layus:
Not sure about what the policy is in HNix about merging my own PR. This seems common for small commits, but I would try to avoid it if possible.
Plus it would assume @expipiplus1 concerns are addressed, and it may not be the case.

Anyone to merge it ?

expipiplus1:
I'm not sure about the policy either, but I'm happy to merge here, or for you to merge on my behalf :)

Is it possible to rebase to get the fixed CI scripts to run though?

layus:
> Is it possible to rebase to get the fixed CI scripts to run though?

Good idea :-)

src/Nix/String.hs
layus: Moved to derivationStrict PR :-)

layus:
I could not resist. I added a commit to get all the tests green :-D. I guess that's the benefit of tests :tada:

expipiplus1:
Thanks!

PR #759, opened 2020-11-17
Bump install-haskell cachix-action and install-nix-action

### LINKED ISSUE #758: github actions need updating
`The `add-path` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution ...`

Should just be a case of bumping the versions of the depended upon actions.

expipiplus1:
Closes https://github.com/haskell-nix/hnix/issues/758

install-nix-action requires an explicit `nix_path` now, do not set this
when it is just used to install nix-instantiate for tests

These are nearly impossible to get right via PRs where they're not all tested :shrug:

### DIFF
--- a/.github/workflows/Cabal-Linux.yml
+++ b/.github/workflows/Cabal-Linux.yml
@@ -29,14 +29,14 @@ jobs:
             ~/.cabal/store
             dist-newstyle
           key: ${{ runner.os }}-Cabal-${{ matrix.ghc }}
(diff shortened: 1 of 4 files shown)


layus (approved):
github actions are indeed a pain to iterate without polluting your git history. But LGTM from eyeballing.

expipiplus1:
Yeah, and CI it better than it was, so probably good to merge

PR #760, opened 2020-11-18
Fix rendering of builtin call frames

layus:
As it happens, we have spent a lot of efforts to bypass the typechecker and get types that are both unused and unconstrained, but still need to  align properly in the end (i.e. renderFrame).

Using exceptions we ensured that there is no way to connect the type of the EvalFrame thrown with the type expected by renderFrame, which is expected to guess it right to print it.

Furthermore, we made it sure to include an EvalFrame constructor that has no use for the v parameter, meaning that we may not even have access to a value of the right type when the constructor is called.

As a result, the code is littered with type annotations forged for the sole purpose of matching two parts of the program disconnected by (untyped) Exceptions.

Examples:

https://github.com/haskell-nix/hnix/blob/38c3062d54e3158e0592e1d25e9b55477049002e/main/Main.hs#L97-L99

https://github.com/haskell-nix/hnix/blob/38c3062d54e3158e0592e1d25e9b55477049002e/src/Nix/Exec.hs#L168

At the places where it works properly, the typing is weird, and still not perfect. It works because `(NValue t f m)` is the only `Scoped` instance we declare.

https://github.com/haskell-nix/hnix/blob/38c3062d54e3158e0592e1d25e9b55477049002e/src/Nix/Eval.hs#L426-L432

This is screaming for typed exceptions, but in the meantime this makes it work :-).

### DIFF
--- a/src/Nix/Exec.hs
+++ b/src/Nix/Exec.hs
@@ -302,7 +302,7 @@ callFunc fun arg = demand fun $ \fun' -> do
       f arg
     NVBuiltin name f -> do
       span <- currentPos
-      withFrame Info (Calling @m @t name span) (f arg)
+      withFrame Info (Calling @m @(NValue t f m) name span) (f arg)
     s@(NVSet m _) | Just f <- M.lookup "__functor" m -> do
       traceM "callFunc:__functor"
       demand f $ (`callFunc` s) >=> (`callFunc` arg)


layus:
See also https://github.com/haskell-nix/hnix/commit/522585a7f1d7d8e72cad5f9a01167c67b694f1af

layus:
@jwiegley I would love to get your feedback on this one. I think it is just a forgotten `s/t/(NValue t f m)/` in 522585a.

layus:
Given that 
1. it does not work ATM
2. the fix makes it working at least in some cases
3. there are some evidences that it could be a single spot missed in a similar rewriting (see above)
4. this project has no strict rule against self merging
5. It has been pending for a while
6. I somehow got commit access ;-)

I am merging this now. It should already be in the previous release :-)

Anton-Latukha:
:smile:

It's a test.

When a person does irrational too abrupt blunt self-merges - loses the merge access.

---

Imagine if here now self merges were not an option.

---

Even I did the project tour over exceptions - Frames are a particularly mysterious part of code for me, I do not know the concept of what "frame" is, and it is some sort of alternative message&exception system which is used only in a couple of places.

Anton-Latukha:
Thanks for the fix to it.

Anton-Latukha:
Thanks for fixing it.

layus:
You can think of frames as a stack of contexts. On error, you can print the stack trace of logical contexts, rather than the Haskell stack trace.

I guess you could think of it as a form of nix evaluation stack maintained for error reporting purposes.

It is implemented with exception values, but these are never thrown AFAIK. A questionable design I guess.

Le 19 décembre 2020 15:54:47 GMT+01:00, Anton Latukha <notifications@github.com> a écrit :

Anton-Latukha:
Yep, pretty much that is how I understood/read the code of it. Tried to refactor/hook-up its exception handling, but found issues doing it, probably you fixed it, because I remember kind of this thing.

Literally today the use of the term, probably more classically, in the DB design of PostgreSQL.

There is also of course a knowledge of Frames in networking Data Links OSI layer.

So the general definition -
Frames essentially are Messages but with a lot of/all supplementary context information of what message is about.

Upon work on exceptions I many times asked how to tie/design the system, and thought to use the Frames, and also they were the only place where I couldn't refactor the internal Exceptions.

So in the end the Frame idea seems legitimate and needs more work, maybe/probably I would rejuvenate the old patch to use the Frame system.

PR #765, opened 2020-12-02
clean-up derivation overrides

Anton-Latukha:
Nixpkgs mainlined these versions.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -223,20 +223,6 @@ let
     root = packageRoot;
 
     overrides = self: super: {
-      # 2020-07-23 hnix uses multiple functions that are unavailable in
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
Strict build fails because of:
```
src/Nix/Expr/Types.hs:179:10: error: [-Wmissing-methods, -Werror=missing-methods]
    • No explicit implementation for
        ‘liftTyped’
    • In the instance declaration for ‘Lift (Fix NExprF)’
    |
179 | instance Lift (Fix NExprF) where
```

So it is not related with the work made, and GHC seems strictified in minor version and started to throw warn for instance, which was previously thrown in 8.10 branch.

PR #766, opened 2020-12-02
CI: GitHub: add Nixpkgs integration test

Anton-Latukha:
This is a tiny project that I want eventually to move into the group.

Details: https://github.com/Anton-Latukha/test-haskell-nixpkgs-integration-action
Implementation gist: https://github.com/Anton-Latukha/test-haskell-nixpkgs-integration-action/blob/main/run-action.sh

This test is to run before release - to know if the project works clean in Nixpkgs, or needs overrides there.

### DIFF
--- a/.github/workflows/Cabal-Linux.yml
+++ b/.github/workflows/Cabal-Linux.yml
@@ -12,7 +12,7 @@ on:
 jobs:
 
   build10:
-    name: GHC
+    name: "GHC"
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


Anton-Latukha:
As expected by the overrides `default.nix` has - `haskellPackages.prettyprinter_1_7_0` override is still needed:
Nixpkgs `haskell-modules/configuration-common.nix`:
```
  hnix = dontCheck (super.hnix.override {
    # 2020-09-18: Those packages are all needed by hnix at versions newer than on stackage
    prettyprinter = self.prettyprinter_1_7_0; # at least 1.7

  });
```

PR #769, opened 2020-12-08
CHANGELOG: explain changes

Anton-Latukha:
Went through the changes merged, it is @layus changes and explained them to users.

[Result](https://github.com/haskell-nix/hnix/blob/2020-12-08-upd-changelog/CHANGELOG.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ## [(diff)](https://github.com/haskell-nix/hnix/compare/0.11.0...master#files_bucket) Progress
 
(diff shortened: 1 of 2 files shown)


layus:
Nice ! Thanks for the meticulous work. I think your updates to shell.nix also deserve a short line here.

Anton-Latukha:
To be realistic, middle-path, to make changelog for the average reader - I try to stick to the changes that are most important for us to advertise and for readers to know - that means info that impacts the interfaces/the use of the library and its changes in functionality that library could automated/relied on.

`shell.nix` is a petty implementation detail of a supplementary code. What also got not mentioned is the reduction of too verbose debug messages you did and CI updates/changes.

Anton-Latukha:
Thanks for correcting the text and adding info on `substring` negative lengths.

PR #775, opened 2020-12-24
clean-up, Values.hs fun code dedup

Anton-Latukha:
Clean-up after "harmonization of the two different value representations" in 522585a7f1d7d8e72cad5f9a01167c67b694f1af


Anton-Latukha:
BTW, looked for a comment like this:
> However, not every term can be reduced to normal form. There are cases where
Nix allows a cycle to exist in the data, so that it can printed simply as
`<CYCLE>`. To represent this, we use a simple recursive type for `NValue`, but
a `Free` construction for `NValueNF`:
>
>    type NValueNF t f m = Free (NValue' t f m) t
>
> The idea here is that `Free` values are those we were able to normalize (since
it has its own terminating base cases of constants, strings, etc), while the
`Pure` thunk is the thunk we'd seen before while normalizing, indicating the
beginning of the cycle.

There was NValue' and NValue and NValueNF, and now NValue "is" NValueNF, because comments in source code were puzzling.

Is NValue is normal form or weak head... :thinking:

Anton-Latukha:
```haskell
normalForm
  :: ( Framed e m
     , MonadThunk t m (NValue t f m)
     , MonadDataErrorContext t f m
     , HasCitations m (NValue t f m) t
     , HasCitations1 m (NValue t f m) f
     , Ord (ThunkId m)
     )
  => NValue t f m
  -> m (NValue t f m)
normalForm = fmap stubCycles . normalizeValue force
```
is a responce.

So `NValueNF` is `normalForm NValue` which may not converge in Nix.

Anton-Latukha:
Probably the explicit imports strategy for refactoring is wrong. Yes, they speed-up the compilation a bit. But the #778 #779 are the right way - in that way there is no need to even restrain the imports to analyze refactoring, what needs refactoring is modules and the exports and the system of exports and modules.

PR #788, opened 2020-12-30
Preparation of release 0.12.0

### LINKED ISSUE #387: Replace all occurrences of `hackyStringMappend` with `principled...` versions.

Anton-Latukha:
What is left is to populate changes introduces in the [Implement derivationStrict primOp](https://github.com/haskell-nix/hnix/pull/554/commits/a36de97df93a50105c045eee4014c1e2e1bec62e) commit.

---

Done.

Rendered [result](https://github.com/haskell-nix/hnix/blob/2020-12-30-release-preparation/CHANGELOG.md).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,17 +1,51 @@
 
 # Changelog
 
-## [(diff)](https://github.com/haskell-nix/hnix/compare/0.11.0...master#files_bucket) Progress
+## [(diff)](https://github.com/haskell-nix/hnix/compare/0.12.0...master#files_bucket) Progress
+
+
+## [(diff)](https://github.com/haskell-nix/hnix/compare/0.11.1...0.12.0#files_bucket) 0.12.0 (2020-12-31)
+
+* Introduction:
+  * New module `Nix.Effects.Derivation`.
+  * Operations on derivations:
+    * old got principled implementations.
+    * also new operations got introduced.
+  * Implementation of the `derivationStrict` primOp.
+
+* Breaking:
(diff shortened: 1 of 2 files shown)


Anton-Latukha:
That took a while to parse.

After all - there are not so many API changes I expected, but there are quite a few.
Most of the work was done in the new module after all.

layus:
Hi,

Thanks for the updates. I would like to make two additions to the release note:

1. Add a warning that most of the public API is there only by accident, and is not to be relied upon.
2. Add a top-level note saying that derivationStrict, and hence every evaluation of a derivation into a store path relies on hnix-store-remote, which means implicitly that it relies on a running nix daemon and will make changes in the real nix store behind.

layus:
The first warning would also simplify the work of providing release notes.

For example, my move to Nix/Effects/Derivation.hs was only meant to separate relatively unrelated code and collect derivation-related functions in a single place. I do not think these functions should be part of any kind of API.

Anton-Latukha:
Yep.

Thank you for the note.

Team (probably means I) need to bite the bullet and move the modules to the `Internal` structure, aka superfluously just graf-lift everything as a preparation phase to the clean-up. So when we clean-up the public interface - we would have distinguished private internal interfaces and public ones and form them properly.

Anton-Latukha:
The note that `derivationStrict` is effectful and how is definitely a must be top.

Anton-Latukha:
I think the same disclaimer probably needs to be in the `readme`.

PR #789, opened 2021-01-01
(needs "Chroot store support" #813) Add an integration test for derivationStrict builtin

Anton-Latukha:
The origianal thread: https://github.com/haskell-nix/hnix/pull/554#issuecomment-752843118

Oversimplification: In this chain Nixpkgs code gets launched and Nixpkgs code runs `fetchurl.nix` file where it is not.

### DIFF
--- a/tests/eval-compare/nixpkgs.hello.nix
+++ b/tests/eval-compare/nixpkgs.hello.nix
@@ -1,0 +1,22 @@
+/*
+
+FIXME: TODO: This is a bad test because:
+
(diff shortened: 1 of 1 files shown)


layus:
This is incredibly complex. The issue is that we run inside a nix chroot store that nix understands, but not hnix. So we do not understand at all that the result path that nix prints neets to be translated to the chroot path.

So, pending on https://github.com/haskell-nix/hnix/issues/813. Alternatively, we could provide our own fetchtarball and ignore the chroot for this test.

PR #793, opened 2021-01-02
hnix.cabal: drop data-dir, prefix data-files instead

### LINKED ISSUE #792: Won't nix build
At the end of the build during installation phase it fails with:

```
Setup: filepath wildcard 'nix/corepkgs/buildenv.nix' does not match any files.
```

sorki:
Fixes
```
Setup: filepath wildcard 'nix/corepkgs/buildenv.nix' does not match any files.
```

Closes #792

### DIFF
--- a/hnix.cabal
+++ b/hnix.cabal
@@ -11,13 +11,12 @@ license:        BSD3
 license-file:   LICENSE
 build-type:     Simple
 cabal-version:  >= 1.10
-data-dir:       data/
 data-files:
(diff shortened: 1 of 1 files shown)


layus:
You should probably also edit https://github.com/layus/hnix/blob/2ec374d19e4a4090abed5513faf6ad6644f0ff8b/src/Nix/Builtins.hs#L313-L326 to add "data" in the nix= path.

PR #794, opened 2021-01-04
Changelog update

Anton-Latukha:
What also needs to be mentioned.
Continuation of: https://github.com/haskell-nix/hnix/pull/788#issuecomment-754188246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,12 +29,13 @@
   * [(link)](https://github.com/haskell-nix/hnix/pull/554/commits/a8e6d28fdb98a1c34f425c8395338fdabe96becc) `Nix.String`: Allow custom computations inside string contexts.
     * By providing `runWithStringContext{T,}'` methods into the API.
   * [(link)](https://github.com/haskell-nix/hnix/pull/554/commits/e45f7632c51a9657f6e8d54c39fd4d21c466d85f) Includded support for new `base16-bytestring`, which advertices 2x-4x speed increase of its operations.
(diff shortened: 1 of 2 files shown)


layus:
Also, I think that #792/#793 should really be merged with the next release. I was unable to reproduce the issue though. @sorki  may have a better idea of how bad this is.

Anton-Latukha:
Well, my view - is we always can print the minor release. Especially in the situation there.

For the Hackage the release works.

The Nixpkgs release is currently hard-blocked in https://github.com/NixOS/nixpkgs/pull/108010 due to https://github.com/haskell-nix/hnix-store/issues/104 which needs all in the https://github.com/commercialhaskell/stackage/issues/5766 be fixed. Wich would take quite a while, especially accounting that a number of the `hnix-store-core` dependencies are not on Stackage at all and that the `cryptohash-sha512` override needs to be smuggled-in. That is all when the `shea` is a maintainer of the Core in the Stackage. So I think that would take a while to sort out, before the Remote release would be mergable into Nixpkgs.

At the same time `sorki` is obscure where and when that error appeared to him, it probably got posted from some development branch he works to fix the situation, because CI `nix-build`s of the `master` does not have that syndrome - they compile and run everything just fine. And if there is indeed the `nix-build` problem - we have a lot of time to fix it until Store `0.4` arrives into Nixpkgs.

I would publish this. The solution to Stackage and to :ship: `0.4` Core&Remote into Nixpkgs is a complex one and can take a month or so, and when they would be solved - there is a reason to merge further fixes and making a minor release specifically to fix some future/particular `nix-builds` for Nixpkgs.

Anton-Latukha:
Fixed the link to the Nixpkgs thread.

PR #799, opened 2021-01-05
default.nix: upd rev & compiler

Anton-Latukha:
After-release Nixpkgs revision update.


Anton-Latukha:
Something needs to be done with these test suites.

The test groups need to be formed, documented, and managed.

And abstain from running some tests when it is `IN_NIX_SHELL` or some `NIX_BUILD*`.

Since Nixpkgs would not happen for quite some time.
Nixpkgs macOS `nix-build` also keeps falling for some time.

PR #801, opened 2021-01-06
CI: Cabal: improvement for the cache, add Haddock & SDist check

Anton-Latukha:
Relates to: https://github.com/haskell-nix/haskell-with-nixpkgs/issues/34

But I'm being lazy and working directly here.


Anton-Latukha:
All works.

PR #804, opened 2021-01-07
WIP: Improvements to make it easier to implement runners other than StandardT

ali-abrar:
* Refactor MonadThunk
* Refactor Nix.Thunk.Basic
* Add Nix.Thunk.Separate
* Add Nix.Fresh.Stable
* Add StableId
* Remove Scopes from Context and replace with ScopeT
* Remove the `t` parameter formerly representing thunks. It can now be inferred from `m`

TODO:
  * [ ] Provide implementations for citations class
  * [ ] Restore Nix.Lint
  * [ ] Fix 3 broken tests
  * [ ] Test thunk refactor more thoroughly (adding new tests as necessary)
  * [ ] Update [Project design intro](https://github.com/haskell-nix/hnix/wiki/Design-of-the-HNix-code-base) in accordance with Thunk changes

Issue #778, opened 2020-12-25
Please, refactor the module grouping


Anton-Latukha:
1. There is some strange not really grouped code around REPL which seems halfway there, some of it was largely taken from Stephen Dehl, and code seems rarely touched and rarely used overall.

For example `Nix.Type.*` modules.

2. Some modules should be renamed.
  * There is `Nix.Expr.Types` and `Nix.Types`, last should be renamed to something like `Hindley-Milner`, which they is/for.
  * There is `Standart`, `Normal`. `Standart` should be better explained/documented. `Normal` is `Normalize`, since it consists of normalization functions, maybe it belongs to `Expr`, maybe to some entity together with `Reduce`
  * `Frames` probably belongs to `Render/`
  * There should be a large group of modules that are `ParserOREval`: `Parser, Exec, Eval, Builtins, Pretty, Fresh, Standard, Normal?, Reduce?, Lint?`

Overall modules around executable/eval are not grouped - they should be. Better grouping allows:
  * to understand design of the project
  * volume of use of modules,
  * people would much easier manage semantics of what modules are about
  * and what modules are closely related,
  * would allows to move towards more modularization & splitting into several smaller viable packages, 
  * so give the devs downstream easier time managing dependencies, 
  * split into several packages localizes the interface change releases.
Something like: `HNix-Types, HNix-Exec, HNix-Extra`, and in Extra put the linting, conversions to/from JSON, XML and so on (JSON and XML do not work and turned off atm.

Anton-Latukha:
https://github.com/haskell-nix/hnix/issues/779#issuecomment-751281485
> I think monorepo + split package could help in conjunction because then people that only need one part are not affected by breaking changes in other parts.

So the very first step - is to group modules on the directory/file level, that would logically form pre-packaging.

Anton-Latukha:
`Atoms` belong to the `Expr`, since exprs are super to atoms. I do not imagine how to use language atoms without language around them (except but reimplementing the language).

Also I think that there is a big benefit to separating between itself the exoskeleton that is an executable implementation and the Nix expression parsing. Expressions are a separate package. Parsing of expressions and what is between executable and the expressions - are in the separate package. And the CLI executable that is an exoskeleton around all that - in the separate package.

That would allow to people to use Expr, or use the parsing logic and code for their own purposes without the code of the executable.

[...]

PR #805, opened 2021-01-07
Children finding their parents

Anton-Latukha:
Aligning instances according to the logical inheritance structure which is layed-out in the #778.

Since this PR is easier, waiting to be mergeable with #804. It changed some of the instances touched here.


Anton-Latukha:
The #804 would free us from the `StdIdT` and so its instances.

Anton-Latukha:
Upstreamed the `instance MonadAtomicRef (ST s)`: https://github.com/mainland/ref-tf/pull/4

PR #807, opened 2021-01-07
CI: GitHub: Nix: dontCheck

Anton-Latukha:
The Cabal testing on 4 GHC versions & bases is really enough.

After that Nix it is really enough just to check compilation - it is the main trouble we have in it.

But `dontCheck` essentially defies the purpose of the Nix dev env, since tests fall in it.

Maybe form a network test switch and disable those tests when the variable is
set. So we configure that in the CI and in Nixpkgs.

Or detect `IN_NIX_SHELL` or build-related env var and disable them automatically.


Anton-Latukha:
macOS build fails due to `hnix-store-*` things, so upstreamed the case.

We need to decide, does Nix dev env in `hnix-store-*` and `hnix` support macOS, or not.

Because the testing needs to go from upstream to downstream.

PR #809, opened 2021-01-07
Simplify Context extension

layus:
Seemed a good idea as it saves some typing when something is added to the context. What do you think ?

### DIFF
--- a/src/Nix/Context.hs
+++ b/src/Nix/Context.hs
@@ -20,16 +20,16 @@ data Context m t = Context
     }
 
 instance Has (Context m t) (Scopes m t) where
-  hasLens f (Context x y z w) = (\x' -> Context x' y z w) <$> f x
+  hasLens f a = (\x -> a { scopes = x }) `fmap` f (scopes a)
 
 instance Has (Context m t) SrcSpan where
-  hasLens f (Context x y z w) = (\y' -> Context x y' z w) <$> f y
+  hasLens f a = (\x -> a { source = x }) `fmap` f (source a)
 
 instance Has (Context m t) Frames where
-  hasLens f (Context x y z w) = (\z' -> Context x y z' w) <$> f z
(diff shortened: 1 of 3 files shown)


src/Nix/Context.hs
```diff
@@ -20,16 +20,16 @@ data Context m t = Context
     }
 
 instance Has (Context m t) (Scopes m t) where
-  hasLens f (Context x y z w) = (\x' -> Context x' y z w) <$> f x
+  hasLens f a = (\x -> a { scopes = x }) `fmap` f (scopes a)
```
Ericson2314: why not `<$>`?

Ericson2314 (approved):
I prefer using field names not positional too.

layus:
@Ericson2314 Do you have any idea if this could trigger a performance drop ?

src/Nix/Context.hs
layus: Because it is pasted from generateLenses :D.

Anton-Latukha: Lerned just yesterday that HLS now has the `lsp-lense-show` feature that allows to peek into the lense generated code.

src/Nix/Options.hs
```diff
@@ -9,7 +9,7 @@ data Options = Options
     , tracing      :: Bool
     , thunks       :: Bool
     , values       :: Bool
-    , scopes       :: Bool
+    , showScopes   :: Bool
```
Anton-Latukha, line 12: @layus 
What do you think, should we rename CLI option from `scopes` to `showScopes`?
Since it is `Enable reporting of scopes in evaluation traces`, I think the `show*` verb is appropriate to describe what option is about, because on itself `--scopes` says nothing of what it really does.
But if we rename it in that fashion - we would also align all other options to this idea.

PR #810, opened 2021-01-07
Add a hacky but working `liftTyped` instance

### LINKED ISSUE #783: Please, implement `liftTyped` for `Lift (Fix NExprF)`
For some time GHC 8.8 strict build started to throw:
```
src/Nix/Expr/Types.hs:179:10: error: [-Wmissing-methods, -Werror=missing-methods]
    • No explicit implementation for
        ‘liftTyped’
    • In the instance declaration for ‘Lift (Fix NExprF)’
    |
179 | instance Lift (Fix NExprF) where
    |          ^^^^^^^^^^^^^^^^^
```

The instance needs implementation, until then - lowering the strictness in #782

layus:
Fixes  #783

### DIFF
--- a/src/Nix/Builtins.hs
+++ b/src/Nix/Builtins.hs
@@ -168,39 +168,7 @@ builtinsList = sequence
   , add0 Normal   "currentSystem"    currentSystem
   , add0 Normal   "currentTime"      currentTime_
   , add2 Normal   "deepSeq"          deepSeq
-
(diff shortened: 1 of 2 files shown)


layus:
We could also consider setting it to `undefined` as we do not use it, and there should be little to no reason to use it anyway.

layus:
This is good enough for now. This is just to fix one warning :-)

PR #815, opened 2021-01-10
Changelog update

Anton-Latukha:


Anton-Latukha:
It is interesting that there is no migration tools.

Was writing the shell script in FP style to help people with migration.

But it becomes obnoxious.

And Haskell scripting needs deps.

So, so far just mentioning the changes.

Issue #767, opened 2020-12-05
Getting a handle on the current state is hard


angerman:
Trying to understand where Hnix is compared to nix. The readme suggests hnix can't evaluate nixpkgs, but it's not very clear what exactly the open issues/complications are. 

If I wanted to replace nix with hnix today, what issues would need to be solved?

sorki:
#554 gets us pretty close - with that PR Hnix produces output paths on par with Nix. That one was blocked by `hnix-store` which was released just a week ago.

Then there's a complicated one which is path filtering (#755 + haskell-nix/hnix-store#72) as it makes our transfomer stacks explode a bit.

That said everything is gonna be slow and we might uncover tons of other issues (Hnix/Nix inconsistencies) when we actually start using Hnix at its full potential.

If you don't need evaluation / building then Hnix is pretty viable and stable solution already and there are few projects that use it as library e.g. 
- https://github.com/expipiplus1/update-nix-fetchgit
- https://github.com/Synthetica9/nix-linter/

angerman:
@sorki thanks! https://github.com/input-output-hk/haskell.nix uses `hnix` as well in https://github.com/input-output-hk/nix-tools, to construct nix expressions. Has been very solid for us. I'm just curious if/when we can take it a step further :-)

My primary motivation for looking at hnix would be nix being too slow. And *really* bad at evaluation, especially around IFDs. I'd like to improve this, but I'd rather not work with C++ if I can help it.

/cc @Ericson2314 -- because you didn't know either :p

layus:
Hi @angerman,

To elaborate on @sorki answer, here is the graps I got while playing with HNix.

At the moment, it is good at parsing nix code. As the parser is feature rich, it can help finding location of expressions, and potentially debug code (but no interface like that exists AFAIK). @sorki has a very good project highlighting what can be done with HNix: https://github.com/sorki/update-nix-file.

With https://github.com/haskell-nix/hnix/pull/554 you get an extra step because you can now evaluate derivations, and get a proper .drv in the nix store. I manged to evaluate nixpkgs.firefox for example. As it is easy to compare the output of nix and hnix, It helped discover a lot of small discrepancies between nix and hnix (https://github.com/haskell-nix/hnix/pull/757, https://github.com/haskell-nix/hnix/pull/756).

[...]

PR #817, opened 2021-01-13
default.nix: upd rev; readme: add Current status section

Anton-Latukha:
"Project status" currently formed by the composition on the theme from https://github.com/haskell-nix/hnix/issues/767#issuecomment-739190263 & https://github.com/haskell-nix/hnix/issues/767#issuecomment-739949611

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,3 +261,8 @@ Currently, the main high-level goal is to be able to evaluate all of Nixpkgs:
 ```shell
 hnix --eval -E "import <nixpkgs> {}" --find
 ```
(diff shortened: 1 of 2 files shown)


Anton-Latukha:
Ok, everything goes whell, "Nix-shell & supplied locall project Hoogle DB" went rebuilding the whole stack, and it should evaluate properly. And this rev update is already an improvement.

Merging it, so would not wait for stack rebuild just to check Hoogle DB, and so I would be able to address the #767.

Issue #777, opened 2020-12-25
Please, make REPL print some useful depth


Anton-Latukha:
Currently, REPL by default is in lazy mode.
Strict mode prints too much.

Normally REPL should print the sane number of levels, something between 2-6.

Also strict probably should use `normalizeValue`, not `normalForm`.

Maybe distinguish thunk and cycke or just replace the `<CYCLE>` with honest `<THUNK>`.

Anton-Latukha:
Previous discussion: https://github.com/haskell-nix/hnix/issues/663

PR #824, opened 2021-01-14
rename '<CYCLE>' to direct & understandable '<expr>'

Anton-Latukha:
Semantically `<CYCLE>` was remotely right only in `opaque` `src/Nix/Normal.hs`
which tries to detect cycles, and only partially right, even if just due to non-termination detenction problem (halting problem).

In other cases, where `<CYCLE>` was used - they are no cycles, which people see: https://github.com/haskell-nix/hnix/issues/663#issue-646681651, and felt confused.

In REPL case - they are simply thunks, or as in XML - just mark expressions.

& they are all expressions after all.

Even most of the infinite nonterminating expressions can be WHNF & studied as expression layer by layer.

And since we can say "expr", that also saves users & devs from communicating what in the lazy FP programming "thunk" slang means.

And currently by itself explains the #777 situation

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ hnix \
   --eval \
   --expr '{ x = true; }'
   
-{ x = "<CYCLE>"; }
+{ x = "<expr>"; }
 ```
 
 To disable laziness add the `--strict` to commands or `:set strict` in the REPL.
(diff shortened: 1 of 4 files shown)


Anton-Latukha:
In Nix the `<CYCLE>` only used in one case:
`data/nix/src/libexpr/eval.cc`
```cpp
static void printValue(std::ostream & str, std::set<const Value *> & active, const Value & v)
{
    checkInterrupt();

    if (active.find(&v) != active.end()) {
        str << "<CYCLE>";
        return;
    }
    ...
}
```

it never appears otherwise and so is not checked in expression tests, so it is a cosmetic output.

Issue #210, opened 2018-04-26
replaceStrings should preserve string context information


jwiegley:
That is, it should propagate the DLists, not drop them as it's now doing.

Permalink:
https://github.com/haskell-nix/hnix/blob/1e2bd9eb2f366da7d98f2aee876bc97f6da37707/src/Nix/Builtins.hs#L507-L535

mightybyte:
I believe this is addressed by #405.

Result:
https://github.com/haskell-nix/hnix/blob/281644a3b2a2b1a488fde771748d88102ab6eb6b/src/Nix/Builtins.hs#L722-L755

Anton-Latukha:
Do the function is long & complex: https://github.com/haskell-nix/hnix/pull/405/files#r240045660

Lets untangle this thread *(makes updates to the messages in the thread)*.

Current code:
https://github.com/haskell-nix/hnix/blob/eb709ccb4a35d4725549914aebfcecc998959750/src/Nix/Builtins.hs#L926-L976

Anton-Latukha:
So:
```
    -- `ns*` goes for NixString here
    fromValue ts >>= \(ns :: NixString) ->
      ...
        go orig result ctx =
          -- and it returns (ctx) OR (ctx <> newCtx)
           ...
      toValue
        $ go (stringIgnoreContext ns) mempty
        $ NixString.getContext ns
```

So I believe that this is indeed closed.

Tail call func `toValue` indeed saves the context, drops the context, and loads new context in place.

PR #828, opened 2021-01-20
Nix.Builtins: replaceStrings: refactor

Anton-Latukha:
After closing the https://github.com/haskell-nix/hnix/issues/210 decided to refactor it.

This is a good example of why, after the project created, a huge volume of such refactors are needed. If before this the code was: https://github.com/haskell-nix/hnix/pull/405/files#r240045660, after refactor - reading the result becomes easy.

And there is seen further refactor possible. And it seen the go is the case of a bit of mutual recursion between `go` and `process`, and looks like they are the same thing.

### DIFF
--- a/src/Nix/Builtins.hs
+++ b/src/Nix/Builtins.hs
@@ -961,25 +961,24 @@ replaceStrings tfrom tto ts =
 
         go orig resultAccum ctx = case lookupPrefix orig of
           Nothing ->
-            maybe
-              (finish resultAccum)
-              (\(h, t) -> go t (resultAccum <> Builder.singleton h))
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
Seem like there is no further room to simplify this.

Anton-Latukha:
I am very solid that this is a pure refactor.

But I want to take this as an opportunity to look into benchmarks of it, and more importantly what testing we do for this (I've seen units) and try to add more rigor to them through some sort of roundtrip generation.

Anton-Latukha:
The Nix replace "" bug prevents an implementation from being elegant and lighter.

PR #857, opened 2021-02-21
Expr.Types: add hashable instances back; organize the module

Anton-Latukha:
They were removed in b5ef442.

Thought to solve the orphans as much as possible, and through having the changelog memo maybe to sort-out the `hashable` situation before release.

Simon did work upstream that solves the root of the question: https://github.com/haskell-unordered-containers/hashable/issues/186, which

`hashable` provided the `NonEmpty` instance, which allows to return them back.

At once, organized the module, now it is easier to understand it altogether. For example, it was not obvious that `Binding` is part of the `NExprF`. And it is also now obvious that project can gain by forming at least 1 submodule from this code.


Anton-Latukha:
Well, since added backward compatibility, the transition to the new `hashable` would be seamless for downstream.

Issue #850, opened 2021-02-11
Maybe reimplement the `demand` and `force` for MonadValue, MonadThunk and such


Anton-Latukha:
Progress:

Proper order the arguments in:
`class MonadThunk t m a | t -> m, t -> a` implementations & all uses of:
  * :heavy_check_mark: `force`
  * :heavy_check_mark: `forceEff`
  * :heavy_check_mark: `further`
  * :heavy_check_mark: `queryM`

`class MonadValue v m`:
  * :heavy_check_mark: `demand`

With #864 factor-out the Kleisli arrows from implementation, and do a new refactor over the code.

  * :heavy_check_mark: `force`
  * :heavy_check_mark: `forceEff`
  * :heavy_check_mark: `further`
  * :heavy_check_mark: `queryM`

`class MonadValue v m`:
  * :heavy_check_mark: `demand`

---

The thunk family of functions are what are the most computationally costly in the system.

`demand` does the `force` (also known as `forseThunk`) - which is one of the most computational-intensive parts of the project:

Currently, they look as:
```haskell
instance ...
  => MonadThunk (StdThunk m) m (StdValue m) where
  force    = force . _stdCited . _stdThunk
  
instance ...
  => MonadValue (Judgment s) (InferT s m) where
  demand = flip ($)
  inform j f = f (pure j)

instance ...
  => MonadValue (StdValue m) m where
  demand (Pure v) f = force v (flip demand f)
  demand (Free v) f = f (Free v)

  inform (Pure t) f = Pure <$> further t f
  inform (Free v) f = Free <$> bindNValue' id (flip inform f) v
```

Notice how thunk gets passed first into the function.

But the thunk (`v`) - is what these functions were created to change, the thunk never gets reused in them.

---

And, as I belabored everywhere already, - the source code does a lot of flipping of the arguments *for them*, and the implementation *of them* flips the arguments internally & to recurse on itself, which suggests that implementation itself should be flipped.

The resulting code:
```haskell
instance ...
  => MonadThunk (StdThunk m) m (StdValue m) where
  force df v = force df (_stdCited $ _stdThunk v)  -- notice

instance ...
  => MonadValue (Judgment s) (InferT s m) where
  demand f v = f v  -- superflous
  inform f v = f $ pure v  -- superfluous
  -- becomes completely superfluous instance, reduction of it preserves code from superfluous type class overhead, simplifies the code and its readability.

instance ...
  => MonadValue (StdValue m) m where
[...]

Issue #861, opened 2021-02-25
Please, form code into canonical thunk locking mechanism


Anton-Latukha:
Currently, the code does the thunk locking/unlocking customly. And customly reacts to the locked thread.

All that should be unified. Unification is simple.

`MonadThunk` needs thunk lock management methods.

Anton-Latukha:
Maybe the handling of the thunks should be hidden. Aka HNix hierarchy allows to abstract MonadThunk instances from the thunk. So just provide the mechanism to the backend as internal top-level functions.

Anton-Latukha:
While the handler pattern that often goes also may be overengineering YAGNI, especially at the current moment.
It probably just needs to be `with...` pattern.

Anton-Latukha:
Essentially was done.

Issue #776, opened 2020-12-24
Please, pack the 'Value.hs' funtions into type classes


Anton-Latukha:
`lmap` is `Profunctor`
`lift` is `MonadTrans`
`unlift` is `MonadTransUnlift`
`hoist` are used only inside the module only to form `lift/unlift`
`bind` is `Monad`
`iter` is `Free`
`sequence` is `Traversable traverse`

Anton-Latukha:
Of course, accounting for the lawful/unlawful implementations and provide tests.

Anton-Latukha:
The RankNTypes use in functions and function context around them, and the cross-types use makes the class formation currently difficult. At least some of.
Needs more tries/looks into it. Maybe constraining back from some higher types where it is now possible.

Anton-Latukha:
If something - it is an "idea to strive for".

I am aware those are not direct `iter` & such, but they are strong HOFs that very close to be a type class instance implementation.
The HOF functions always can be given default arguments (like `id` for natural transformations) & be specialized to particular instance implementation.

Due to the design of HNix & how already these HOFs are close to be instances - at some point, it is possible to "suddenly" get them as "weakly lawful" or lawful instances.

Some functions (currently) do not align by types to be for example `iter`, as `iterNValue` operates on thunks and transfers its type to `NValue` - so its signature does not align to `iter` completely.

Thou from the list there can be cases where implementation gets aligned with famous type classes.

Maybe it is a good idea to create equivalent type classes for those types of functions if we would have several sets of those for different types (aka for `N{Expr,Thunk,Value}`).

Anton-Latukha:
With the number of funtions there - it would be ugly, but the functions should be duplicated & useless argument passing removed from them (as it was done in `MonadThunk`)- that would bring functions closer to type classes. This requires project-wide refactoring.

PR #862, opened 2021-02-26
Flipping `demand`

Anton-Latukha:
A milestone stage towards #850

Remember that `force` would lose the functional argument.
`demand` would be something along the lines of `free -> f =<< force thunk`.

By the result of the `demand` - it would be seen, is in viable to keep it, or simply replace with that `free -> f =<< force thunk` or move `free` logic into `force` and use `force` with `=<<`.

So instead of the `demand` it would be between function and its argument would be `force v`.

In other words, unflipping required to unfold the code.

But the further following folded elegantness worth it.

Also all those `\case` really are helper functions that explain themself.

Or, `alg` if you want.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -17,6 +17,8 @@
       * `querryM`
       * `querryThunk`
 
+  * [(link)](https://github.com/haskell-nix/hnix/pull/862/files#diff-caa5d6592de00a0b23b2996143181d5cb60ebe00abcd0ba39b271caa764aa086) `Nix.Value.Monad`: `class MonadValue v m`: `demand` unflipped the arguments. All its implementations got more straigh-forward to use and `demand` now tail recurse.
+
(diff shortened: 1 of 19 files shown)


Anton-Latukha:
Currently surprised the benchmark saves are still small.

Sadly, the benchmarks so far not show huge improvement.

But need to inspect the profiling more.

I wait on one at the end of #850.

But more and more thinking that 1-threadness with no system of lock management, flipped style in which the project was created produces unoptimized code and lazy boxed types are the main points. Maybe they are the main nails performance hangs on in the thunk processing.

The more we streamline the code - the more magnitude of the gains would start to snowball.

Anton-Latukha:
Awakened the swarm nest for project improvements:

#864
#861
#850
#776

, and others.

Checked for it to eval Nixpkgs in the same way. During evaluations found several typos, until evaluations became same.

PR #867, opened 2021-03-01
Builtins: refactor; Utils: inline what is definitely inlinable; fx to Eq1 NValue'; fx all warns

Anton-Latukha:
Code is literally the code from the `base: Data.Bool`, in there it is also does not
have inline directive:
```haskell
-- | From @Data.Bool ( bool )@.
bool :: a -> a -> Bool -> a
bool f _ False = f
bool _ t True  = t
```
Through the profiler noticed that `bool` was drawing resources.

GHC was not inlining the `bool`. Imagine that.

So we should probably carefully start giving the nudges to the GHC to inline simple non-recursive code.

These Utils functions are a definite start.

This actually seems to shave-off quite well.

### DIFF
--- a/src/Nix/Value.hs
+++ b/src/Nix/Value.hs
@@ -21,12 +21,19 @@
 module Nix.Value
 where
 
-import           Control.Comonad                ( Comonad, extract )
+import           Control.Comonad                ( Comonad
+                                                , extract
+                                                )
 import           Control.Exception              ( Exception )
 import           Control.Monad                  ( (<=<) )
 import           Control.Monad.Free             ( Free(..)
-                                                , hoistFree, iter, iterM )
-import           Control.Monad.Trans.Class      ( MonadTrans, lift )
+                                                , hoistFree
+                                                , iter
+                                                , iterM
+                                                )
+import           Control.Monad.Trans.Class      ( MonadTrans
+                                                , lift
+                                                )
 import qualified Data.Aeson                    as Aeson
 import           Data.Functor.Classes           ( Show1
                                                 , liftShowsPrec
@@ -38,7 +45,8 @@ import           Data.Typeable                  ( Typeable )
 import           GHC.Generics                   ( Generic )
 import           Lens.Family2.Stock             ( _1 )
 import           Lens.Family2.TH                ( makeTraversals
-                                                , makeLenses )
+                                                , makeLenses
+                                                )
 import           Nix.Atoms
 import           Nix.Expr.Types
 import           Nix.Expr.Types.Annotated
@@ -187,15 +195,22 @@ hoistNValueF
   :: (forall x . m x -> n x)
   -> NValueF p m a
   -> NValueF p n a
-hoistNValueF lft = \case
-  NVConstantF a  -> NVConstantF a
-  NVStrF      s  -> NVStrF s
-  NVPathF     p  -> NVPathF p
-  NVListF     l  -> NVListF l
-  NVSetF     s p -> NVSetF s p
-  NVClosureF p g -> NVClosureF p (lft . g)
-  NVBuiltinF s g -> NVBuiltinF s (lft . g)
-
+hoistNValueF lft =
+  \case
+    -- Pass-through the:
+    --   [ NVConstantF a
+    --   , NVStrF s
+    --   , NVPathF p
+    --   , NVListF l
+    --   , NVSetF s p
+    --   ]
+    NVConstantF a  -> NVConstantF a
+    NVStrF      s  -> NVStrF s
+    NVPathF     p  -> NVPathF p
+    NVListF     l  -> NVListF l
+    NVSetF     s p -> NVSetF s p
+    NVBuiltinF s g -> NVBuiltinF s (lft . g)
+    NVClosureF p g -> NVClosureF p (lft . g)
 {-# inline hoistNValueF #-}
 
 -- * @__NValue'__@: forming the (F(A))
@@ -625,40 +640,43 @@ data ValueType
 
 -- | Determine type of a value
 valueType :: NValueF a m r -> ValueType
-valueType = \case
-  NVConstantF a -> case a of
-    NURI   _ -> TString NoContext
-    NInt   _ -> TInt
-    NFloat _ -> TFloat
-    NBool  _ -> TBool
-    NNull    -> TNull
-  NVStrF ns  ->
-    TString $
-      bool
-        NoContext
-        HasContext
-        $ stringHasContext ns
-  NVListF{}    -> TList
-  NVSetF{}     -> TSet
-  NVClosureF{} -> TClosure
-  NVPathF{}    -> TPath
-  NVBuiltinF{} -> TBuiltin
+valueType =
+  \case
+    NVConstantF a ->
+      case a of
+        NURI   _ -> TString NoContext
+        NInt   _ -> TInt
+        NFloat _ -> TFloat
+        NBool  _ -> TBool
+        NNull    -> TNull
+    NVStrF ns  ->
+      TString $
+        bool
+          NoContext
+          HasContext
+          (stringHasContext ns)
+    NVListF{}    -> TList
+    NVSetF{}     -> TSet
+    NVClosureF{} -> TClosure
+    NVPathF{}    -> TPath
+    NVBuiltinF{} -> TBuiltin
 
 
 -- | Describe type value
 describeValue :: ValueType -> String
-describeValue = \case
-  TInt               -> "an integer"
-  TFloat             -> "a float"
-  TBool              -> "a boolean"
-  TNull              -> "a null"
-  TString NoContext  -> "a string"
-  TString HasContext -> "a string with context"
-  TList              -> "a list"
-  TSet               -> "an attr set"
-  TClosure           -> "a function"
-  TPath              -> "a path"
-  TBuiltin           -> "a builtin function"
+describeValue =
+  \case
+    TInt               -> "an integer"
+    TFloat             -> "a float"
+    TBool              -> "a boolean"
+    TNull              -> "a null"
+    TString NoContext  -> "a string"
+    TString HasContext -> "a string with context"
+    TList              -> "a list"
+    TSet               -> "an attr set"
+    TClosure           -> "a function"
+    TPath              -> "a path"
+    TBuiltin           -> "a builtin function"
 
 
 showValueType :: (MonadThunk t m (NValue t f m), Comonad f)
--- a/src/Nix/Value.hs
+++ b/src/Nix/Value.hs
@@ -196,6 +196,7 @@ hoistNValueF lft = \case
   NVClosureF p g -> NVClosureF p (lft . g)
   NVBuiltinF s g -> NVBuiltinF s (lft . g)
 
+{-# inline hoistNValueF #-}
 
 -- * @__NValue'__@: forming the (F(A))
 
@@ -266,6 +267,7 @@ iterNValue'
   -> r
 iterNValue' k f = f . fmap (\a -> k a (iterNValue' k f))
 
+-- *** Utils
 
 -- | @hoistFree@: Back & forth hoisting in the monad stack
 hoistNValue'
@@ -276,7 +278,7 @@ hoistNValue'
   -> NValue' t f n a
 hoistNValue' run lft (NValue v) =
     NValue $ lmapNValueF (hoistNValue lft run) . hoistNValueF lft <$> v
-
+{-# inline hoistNValue' #-}
 
 -- ** Monad
 
@@ -455,6 +457,7 @@ iterNValueM transform k f =
     go (Pure x) = Pure <$> x
     go (Free fa) = Free <$> bindNValue' transform go fa
 
+-- *** Utils
 
 -- | @hoistFree@, Back & forth hoisting in the monad stack
 hoistNValue
@@ -464,7 +467,7 @@ hoistNValue
   -> NValue t f m
   -> NValue t f n
 hoistNValue run lft = hoistFree (hoistNValue' run lft)
-
+{-# inline hoistNValue #-}
 
 -- ** MonadTrans
 
--- a/src/Nix/Value.hs
+++ b/src/Nix/Value.hs
@@ -53,6 +53,7 @@ import           Nix.Expr.Types.Annotated
 import           Nix.String
 import           Nix.Thunk
 import           Nix.Utils
+import           Data.Eq.Deriving
 
 
 -- * @__NValueF__@: Base functor
@@ -223,7 +224,7 @@ newtype NValue' t f m a =
     -- | Applying F-algebra carrier (@NValue@) to the F-algebra Base functor data type (@NValueF@), forming the \( F(A)-> A \)).
     _nValue :: f (NValueF (NValue t f m) m a)
     }
-  deriving (Generic, Typeable, Functor, Foldable, Eq1)
+  deriving (Generic, Typeable, Functor, Foldable)
 
 instance (Comonad f, Show a) => Show (NValue' t f m a) where
   show (NValue (extract -> v)) = show v
@@ -617,6 +618,7 @@ pattern NVClosure x f <- Free (NVClosure' x f)
 pattern NVBuiltin name f <- Free (NVBuiltin' name f)
 
 
+
 -- * @TStringContext@
 
 data TStringContext = NoContext | HasContext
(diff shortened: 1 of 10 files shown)


src/Nix/Value.hs
```diff
@@ -622,40 +640,43 @@ data ValueType
...
-  TNull              -> "a null"
-  TString NoContext  -> "a string"
-  TString HasContext -> "a string with context"
-  TList              -> "a list"
-  TSet               -> "an attr set"
-  TClosure           -> "a function"
-  TPath              -> "a path"
-  TBuiltin           -> "a builtin function"
+describeValue =
+  \case
```
Ericson2314, line 670: I've never seen this style. It would seem to be extra indent for no purpose?

Ericson2314:
We always do production builds with `-fexpose-all-unfoldings` which gives GHC the option to inline everything. (That's also basically how rustc works, FWIW.) I do think that's easier than worrying about this stuff with individual INLINE or INLINEABLE pragmas.

src/Nix/Value.hs
Anton-Latukha: Well, it is a style called diagonal indentation.

The idea is that all the code indeed seems seamless, where the function is underlined by the argument it accepts.

Since Haskell generally is designed to be read from right to left, the expressions generally should go in the same order. AKA `>>=` is left to right, that is why it reads hard in the right to left expression.

Isn't it can be persived weird that function happens to be on the right and its arguments are on the new line left. For example:
```haskell
attrSetAlter (k : ks) pos m p val = case M.lookup k m of
  Nothing | null ks   -> go
          | otherwise -> recurse M.empty M.empty
  Just x
    | null ks
    -> go
    | otherwise
    -> x >>= fromValue @(AttrSet v, AttrSet SourcePos) >>= \(st, sp) ->
      recurse (demand pure <$> st) sp
      
-- or

    pushWeakScope ?? body $ demand s $ unpackSymbolic >=> \case
      NMany [TSet (Just s')] -> pure s'
      NMany [TSet Nothing] -> error "NYI: with unknown"
      _ -> throwError $ ErrorCall "scope must be a set in with statement"
    pushWeakScope ?? body $
```

Isn't that strange that pattern matching shows-up on the left, but the head of pattern match structure is completely in other place, and it is not indented with the pattern match in any way?

It is for pattern-match, but we do this with function definitions that have any function as a head. Somehow we like to provide the head on the right, and provide arguments to it on the left.  Its arguments suddenly appear across the newline in opposite direction.

In other words that visually does not create a structure between what is matched to what, or these arguments are to which function.

Diagonal style code reads from the tail to the head and diagonally from the end to start. Uses more space - but provides that structuring.

So, because language goes from the end of the expression to the beginning - diagonal alignment shows what are the layers.

Also all code structures then follow more uniform, structure. With `bool`, `maybe` `either` and others - especially.

The idea (I have), is that Lisp is the ease refactorable language.
Haskell is harder to refactor, at large because it obstructs the lambdas by hard-syntax.

The declarative code is bit-by-bit seems more understandable. And it easier to develop in it. But it is noisy, there is a lot of repetitiveness of the same patters in it, matches to `Just` are just the beginning, because it is more structurally noisy and requires more repetitiveness, in some declarative structures it is not possible to factor-out the tail or the head of the expressions:
```haskell
function Pattern match = pure this
function New match     = pure that
function otherwise        = pure nothing
```
How one refactors it to dedupe the `pure`s without creating a new function or using lispy lambda things.

With a more heavy declarative style it gets harder and harder for people to see the patterns to refactoring in it.

For example:
```haskell
wrapParens :: OperatorInfo -> NixDoc ann -> Doc ann
wrapParens op sub
  | precedence (rootOp sub) < precedence op
  = withoutParens sub
  | precedence (rootOp sub)
    == precedence op
    && associativity (rootOp sub)
    == associativity op
    && associativity op
    /= NAssocNone
  = withoutParens sub
  | otherwise
  = parens $ withoutParens sub
```

It is already hard to see what is going on there.

Moreover, notice that the two sequential matches are indeed one parallel "or".

Result is:

```haskell
wrapParens op sub =
  bool
    parens
    id
    (precedence (rootOp sub)       < precedence op
    || (precedence (rootOp sub)   == precedence op
        && associativity (rootOp sub) == associativity op
        && associativity op /= NAssocNone)
    )
    (withoutParens sub)
```

We always do `(withoutParens sub)`, but are parens are added - depends on the bool condition. In fact `id` is just `mempty` for functions, and so the `whenBase`, `whenProper` can be used for all such binary constructor types:
```haskell
wrapParens op sub =
  whenBase
    parens
    (precedence (rootOp sub)       < precedence op
    || (precedence (rootOp sub)   == precedence op
        && associativity (rootOp sub) == associativity op
        && associativity op /= NAssocNone)
    )
    (withoutParens sub)
```
- the special case is the stub and the base case adds `parens`. Well, `==` are really `/=` match for the compiler, but inversion would be harder to read for people.

It can be to use just 1 function instead of type-specific bool, `maybe`, `either`, `list`, `free` - all binary constructor structures, and maybe also structures that have a base case constructor. and 1 function for a special case that returns a stub in the base case, and 1 function for the base case that returns a stub in the special case. That Haskell is not able to do it - is Haskell type system specifics, it can not `id -> mempty` in a huge portion of cases, which rules-out the last two for when the function gets passed-through, and these functions have type class overhead jumps. People previously tried to do it, at least several times I know of, but `id` complexities and type class jump prevented people for giving *this type of library. Maybe because those people tried to do that before they learned advanced stuff, because as we see - such simple switcher things can be inlined.

---

And:
```haskell
attrSetAlter (k : ks) pos m p val = case M.lookup k m of
  Nothing | null ks   -> go
          | otherwise -> recurse M.empty M.empty
  Just x
    | null ks
    -> go
    | otherwise
    -> x >>= fromValue @(AttrSet v, AttrSet SourcePos) >>= \(st, sp) ->
      recurse (demand pure <$> st) sp
```

Becomes:

```haskell
attrSetAlter (k : ks) pos m p val =
  bool
    go
    (maybe
      (recurse M.empty M.empty)
      (\x ->
        do
          (st, sp) <- fromValue @(AttrSet v, AttrSet SourcePos) =<< x
          recurse (demand pure <$> st) sp
      )
      (M.lookup k m)
    )
    (not $ null ks)
```

Now for the person and probably for the GHC it is now more obvious, that there is no need to do `M.lookup k m`, if it is simply `(null ks)`.


---

It would be strange to break things in the working project or after having dirty migration into sub-packages.

This year the reactors is what I care about. The project was created but never saw a big refactor work. Before proclaiming the working base, for future effectiveness, we need to squeeze-in all possible refactors, changes, and migrations before proclaiming the MVP release and before splitting into sub-packages.

There are essentially 3 migrations this year. 1. clean-up and specialization of the code, 2. reorganizing the modules (there is a lot of the top-level modules that need to be grouped into subgroups so in the future to making sub-packages from those subgroups.

---

I found that once declarative code is moved into lambdas with diagonal style - the patterns show up by themselves. And so it is easier to see what to refactor. They are in Lisp - so it is easier to refactor them. And the diagonal lispy style is the easiest to restructure and refactor.

And as core language syntax bleeds into DSL, System F bleeds its syntax in a Haskell, so it is quite easy and powerful to write in a Lisp-y style in Haskell.

The answer is - it is easy to refactor in it and Haskell System F undercurrent language design ephemerally aligns with it.

And the project really needs refactoring. I would debate that lispy style and diagonal alignment is not stranger than the unclassical unoptimized function argument order and code style trying aligning Haskell control flow from head to tail (when it anyway starts executing from the tail, jumps to the head, then jumps closer into the tail, because functions go `<-<-`, written monad binds go `>>=` argument does `<&>` jump into a function with `??` the `demand t f` which passes though `f` into `force t f`, which simply applies it at the very end. And my (and other readers, GHC probably) head goes doing `⮒` which is probably a fold left. It looks like trying to make Haskell do C style. But Haskell anyway designed to execute from the tail to end, if not anything - as lazy language), which style was used during creation a lot. It is understandable why, but the creation phase does not mean perfect code. Refactoring is what cleans-up the implementation and shows the implementation elegance.

And my perception is that so far I like this style very much, it makes the reading and work very-very seamless, shows the logic and is very powerful, especially to refactor in. It is strange that people shun away from it and not employ it much. Maybe it is not more prevalent because no one pays for Lisp or refactoring.

In the end, if the things are better looking in `do` notation - I'm wrapping things in `do` notation. When the lispy Haskell features fall short to what declarative gained - using declarative there.

So the answer is that the style indeed depends.

Anton-Latukha: The lispy-type of code is even more powerful to refactor because of the Structured Haskell Mode.

Structured programming is Lisp-born and is powerful in it, because, as known, of ap parens reasons.

But in my case, the Structured Haskell Mode does not work in my Emacs (VI bindings have a side-car binding system, SHM uses native binding system).

Ericson2314: I will have to finish reading this later. I do have a soft spot for "Structured Haskell Mode". I would not recommend replacing pattern matching with `maybe` and `bool` though in large cases, however. Even Lisp came towards pattern matching with https://docs.racket-lang.org/guide/syntax-case.html and https://www.gnu.org/software/guile/manual/html_node/Syntax-Case.html. https://www.pyret.org/ led by one of the https://htdp.org/ authors also makes pattern matching part of the language.

Bottom line is yes there are things to learn from lisp and structured editing is very good, but I would not make avoiding pattern matching one of those lessons.

Anton-Latukha:
> We always do production builds with `-fexpose-all-unfoldings` which gives GHC the option to inline everything. (That's also basically how rustc works, FWIW.) I do think that's easier than worrying about this stuff with individual INLINE or INLINEABLE pragmas.

Yes.

Just gradually tipping toes.
If we decide - reduction of all `^.*{-# INLINE .*#-}$` statements is one regexp commit away.

---

HNix uses recursivity a lot. Both in functions and in `Fix (Fix)` types, and GHC or binary size can (going to) blow-up inlining the recursive things. So being careful with inlining and aggressive specialization of `fmap` on some `Fix (Fix)`.

If GHC just failed to inline default `base` `bool` implementation in the default settings...

---

Cabal description already has `--optimize` that `ghc-options: -fexpose-all-unfoldings -fspecialise-aggressively -O2` but that is when one has a hammer everything looks specialized by unfolds.

But with recursivity and unrefactored code - do not think it would give the results we are hoping for.

---

The code clean-up, and module refactoring, reducing RankNTypes after the hacking of the recursive type system togather, hiding internal things in the internal module namespace increase the magnitude of all kinds of GHC optimizations.

Indeed, it just needs to be determined where:
```haskell
{-# OPTIONS_GHC -fexpose-all-unfoldings #-}
{-# OPTIONS_GHC -fspecialise-aggressively #-}
```
Should be put module-wise.

Want to refactor the carrier (thunks) as first option and Kliesli arrows from last options away from the main control path first, before asking the compiler aggressively optimize, otherwise, it is being like "we have not optimized it, so maybe you would optimize it for us?", the answer for the HOF-heavy complex code going to be the more complex - the more GHC would say "nope".

I want to at least take-out an `f` from `demand f t = (demand f) =<< force t`, before looking into GHC options.

src/Nix/Value.hs
Anton-Latukha: It all started with someone somewhere in the deep past suggesting to try to use `maybe` and stick to it to open a forgotten realm.

Anton-Latukha: Sorry, maybe it was a bit direct rambling.

> I've never seen this style. It would seem to be extra indent for no purpose?

They serve the purpose.
All the heads of the functions are at the beginning of the lines. And as such inclusion of 1 function into another - shown by the indentation level.

Code reads diagonally from down right to the up left, or from the up-left to the down-right. Ahe reader at once sees all the functions and what they carry, eye movement ends at the function head function and the function definition name.

That is the purpose of the indentation.

I also started to like indenting the `do` sometimes, to diagonally outline the `do` blocks inside the function body by indentation.

`\case` is a lambda function, with `\case` being the head of the function, and as such - gets indented.
So diagonal indentations have a purpose.

Haskell being the abstract terse language can bear the diagonal alignment of code. If the Haskell function does not fit on the screen it needs to be refactored, if that is not just a large pattern match to constructors.

So those indented `\case`, `case` `\ ..`, when they are being more than 2 lines - are just a nod as an already aligned function in the code waiting to be just properly named local function, so style nudges the people to explain it with local function verbs that also make the code more easily readable.
It is a Lisp style, that enables refactoring in the code. And allows to untangle the code that was written, not only show the possible refactors in it but also refactor it at the same time.

Please, bear with me, yes, the style sometimes can look obnoxious, but it does the job really well and it is the style that produces more and more organization and elegance at every passage over it.

Anton-Latukha: Good example. how in the head to parse this:
```haskell
mapAttrs_ f xs = fromValue @(AttrSet (NValue t f m)) xs >>= \aset -> do
  let pairs = M.toList aset
  values <- for pairs $ \(key, value) ->
    defer @(NValue t f m)
      $   withFrame Debug (ErrorCall "While applying f in mapAttrs:\n")
      $   callFunc ?? value
      =<< callFunc f (nvStr (makeNixStringWithoutContext key))
  toValue . M.fromList . zip (fmap fst pairs) $ values
```
???

It indented good - yes.
But we see here that the terse & pretty indentation is not a pinnacle of the code.
Is this indented code is, in fact, parsable, by a person, that programs the project?

How the `$` `??` `=<<` map where what?

To understand it - the person going to naturally start sectioning the code with parens. Aka looking where the code sections, aka splitting this code into lambdas. Aka breaking this style down into Lisp.

Anton-Latukha:
```haskell
mapAttrs_ f xs =
  do
    aset <- fromValue @(AttrSet (NValue t f m)) xs
    let
      pairs = M.toList aset

    values <-
      (traverse
        (\ (key, value) ->
          (defer @(NValue t f m)
            (withFrame Debug (ErrorCall "While applying f in mapAttrs:\n")
            ((=<<)
              (\ x -> (callFunc ?? value) x)
              (callFunc f (nvStr (makeNixStringWithoutContext key)))))))
        pairs
      )
    (toValue (M.fromList (zip (fmap fst pairs)))) values
```

Of course, it also can be:
```haskell
mapAttrs_ f xs = do
    nixAttrset <- fromValue @(AttrSet (NValue t f m)) xs
    let
      keyVals = M.toList nixAttrset
      applyFunToKeyVal (key, val) = do
        runFunForKey <- callFunc f $ nvStr $ makeNixStringWithoutContext key
        callFunc runFunForKey val
    newVals <- traverse (defer @(NValue t f m) .
      withFrame Debug (ErrorCall "While applying f in mapAttrs:\n") . applyFunToKeyVal
      ) keyVals
    toValue $ M.fromList $ zip (fst <$> keyVals) newVals
```

This a resulting code with the empty lines removed.

But next tend to read better:
```haskell
mapAttrs_ f xs =
  do
    nixAttrset <- fromValue @(AttrSet (NValue t f m)) xs
    let
      keyVals = M.toList nixAttrset
      keys = fst <$> keyVals

      applyFunToKeyVal (key, val) =
        do
          runFunForKey <- callFunc f $ nvStr $ makeNixStringWithoutContext key
          callFunc runFunForKey val

    newVals <-
      traverse
        (defer @(NValue t f m) . withFrame Debug (ErrorCall "While applying f in mapAttrs:\n") . applyFunToKeyVal)
        keyVals

    toValue $ M.fromList $ zip keys newVals
```

There are scientific papers that determined that for readability the most important alignment property is the amount of space between lines.
People understand the text better when there is more space between lines, it is psychologically easier to read the text.

The best open sum-up material that I found on it so far is: https://ajet.org.au/index.php/AJET/article/view/2286/1117. It is an old studied scientific topic, so there are no new open studies or papers on it. In there is a number of references to the interconnection between readability, typography, and linespace. In short, good linespace helps readability a lot.

Anton-Latukha: > I will have to finish reading this later. I do have a soft spot for "Structured Haskell Mode". I would not recommend replacing pattern matching with maybe and bool though in large cases, however. Even Lisp came towards pattern matching...

Of course. `maybe` and such can not replace complex pattern matches, `bool` `maybe` `either` are binary, while pattern matches are multicase.
Also Haskell pattern matches go together with guards, ViewPatterns and such others.

I so far not once replaced the complex pattern match in the code, you can check 8). There is only 1 very minor `null` dup sin I just contributed, but the dup there replicates the upstream Nix bug behavior.

Anton-Latukha: +===

Well, I would retreat again.

I looked at the line marked in the review message, and at the message. And knowing the style I wrote PR in, started to discuss it. But not looked what the code is beneath the line marked. Beneath the marked text, alignment of those case statements, yes, alignment there is a bit vacuous.

Just during the work process aligned these also. If possible to allow by the align rules, tend to favor giving more align-right to the biggish functions and the biggish pattern matches,  to indicate to the eyes that they are not at the beginning of the line and that they are in fact a block of code inside something that is up the screen.

But that is a minor thing for me, maybe would try to obstrain more from that, also can revert these.

PR #868, opened 2021-03-03
default.nix: hnix-store-core became normal version

Anton-Latukha:
We pulled back the `0.2` from Stackage Nightly & in future LTS. So the `hnix-store-core` processes returning to normal.

Thou, these very recent upstream changes somehow managed to break the old reproducible build (we have Nixpkgs pinned), which is amusing in itself and suspicious at the same time.
So because of that, I met the change while doing Haskell work in #867.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -91,20 +91,20 @@
 #   , nixos-20.03  # Last stable release, gets almost no updates to recipes, gets only required backports
 #   ...
 #   }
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
Well, CI speaks for itself.

It is strange that reproducible build broke. Our `if then else` for `rev` looks normal, IDK so far.

PR #873, opened 2021-03-04
Migrating code to new demand implementation

Anton-Latukha:
- :heavy_check_mark: Gradual switches where posible.
 - :heavy_check_mark: All uses of it, including all `Builtins` are processed.
 - :heavy_check_mark: A couplr of several level `do` blocks became 1.
 - :heavy_check_mark: Several of the monadic binds became functors.
 - :heavy_check_mark: ChangeLog (also would be rewritten couple of times by the further change updates)
 
This almost fully closes the https://github.com/haskell-nix/hnix/issues/850

What is further left there is basically to move Kleisli out of `inform`, `informF` is for that, then do the includes of the function uses inside the `do` blocks and fold the lambdas, binds and `do` blocks further. With the current lispy sectioning, it is easy and semi-automatic.

### DIFF
--- a/src/Nix/Effects/Derivation.hs
+++ b/src/Nix/Effects/Derivation.hs
@@ -351,7 +351,7 @@ buildDerivationWithContext drvAttrs = do
     -- common functions, lifted to WithStringContextT
 
     demandF' :: (NValue t f m -> WithStringContextT m a) -> NValue t f m -> WithStringContextT m a
-    demandF' f v = join $ lift $ demandF (pure . f) v
+    demandF' f v = join $ lift $ f <$> demand v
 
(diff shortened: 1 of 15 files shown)


src/Nix/Effects/Derivation.hs
```diff
@@ -351,7 +351,7 @@ buildDerivationWithContext drvAttrs = do
     -- common functions, lifted to WithStringContextT
 
     demandF' :: (NValue t f m -> WithStringContextT m a) -> NValue t f m -> WithStringContextT m a
-    demandF' f v = join $ lift $ demandF (pure . f) v
+    demandF' f v = join $ lift $ f <$> demand v
```
layus, line 354: Note to myself: It seems that demand' -> demandF' was not reverted. For consistency it should probably also be flipped.

layus: https://github.com/haskell-nix/hnix/pull/876

PR #876, opened 2021-03-05
refactor demand' with new style

layus:
@Anton-Latukha I did not realize (until now) that you completely removed the `f` parameter to `demand`. Nice job !

In this PR I also reverted the introduction of `bool`, way to subject to confusion as there is no way to tell if the first argument is the True or False case from the type. (This does not happen with other such "deconstructors", such as `maybe` used right above, because the `Just` case take an argument.)

### DIFF
--- a/src/Nix/Effects/Derivation.hs
+++ b/src/Nix/Effects/Derivation.hs
@@ -318,18 +318,13 @@ buildDerivationWithContext drvAttrs = do
           mHash
 
       -- filter out null values if needed.
-      attrs <-
-        bool
-          (pure drvAttrs)
-          (M.mapMaybe id <$> forM drvAttrs
-            (demandF'
-              (pure . \case
-                NVConstant NNull -> Nothing
-                value -> pure value
-              )
-            )
+      attrs <- lift $ if not ignoreNulls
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
But that is the idea.
The `bool`, `maybe` `either`, etc have the same form. The first argument - is the base case, and it is `False` for the Bool, as Nothing for the Maybe, as Left for the Either.

In `either` it is also not possible to tell which argument is `Left a`, in `free` which argument is `Pure a`, until the person understands that the first argument is the base case in all of them, it is a small knowledge to be able to use and read them all.

So the idea is that `bool` provides the base case first. As I found it is easier to flip the flag aka `(not $ null xs)` to move the base case in the proper place.

It is often found that the base case often a function, a special case just augments that function, so the function moves-out and produces `mempty` or `id`, and another part of the cases the base case is `mempty` or `pure` or `id`, so than even further can be literally said: in the base case is a stub, in the special - this is applied.

And all these functions make most of the code switchers have a uniform structure, which allows refactoring and templating automatic changes on the code structure.

Just explaining why started using them.
Would not argue for others to change them back.

PR #878, opened 2021-03-08
unflip {mkNixDoc, nvSet{,'}}; refactors

Anton-Latukha:


src/Nix/Eval.hs
```diff
@@ -499,9 +509,14 @@ addStackFrames
...
-  scopes <- currentScopes :: m (Scopes m v)
-  withFrame Info (EvaluatingExpr scopes v) (f v)
+addStackFrames f v =
+  do
+    scopes <- currentScopes :: m (Scopes m v)
+
+    -- sectioning gives GHC optimization
+    (`withFrameInfo` f v) $ (`EvaluatingExpr` v) scopes
+ where
+  withFrameInfo = withFrame Info
```
layus, line 523: I am trying to understand how to optimize with GHC. Would you mind explaining how this optimisation works ?

Anton-Latukha: Well, I simply give a semantic nudge to the GHC.

I simply compared the CPU cycles between profiling results, it was 2.4% before vs ~1.6-8% in several runs afterward.

Since the second arguments are more static here - they should be first arguments and grouping should indicate that there is a direct static grouping of functions with `{f, v}` which are in the definition signature.

I just imagine how GHC parses the source code and detects that `( .. )` block - is static, because `('withFrameInfo' f v)` is static for `addStackFrames f v =`.  And the `scopes` is a value that calculates in the monadic context and depends on the monadic context, it dissects the `v` into `[Binding (m v)]` & runs `currentScopes :: m (Scopes m v)`, so the `v` does not pass-through there - it gets dissected, which is hard to optimize then simple static pass-through reuse, so the `scopes` is the tail.

Simply comparing:
```haskell
addStackFrames f v =
    
  withFrame Info (EvaluatingExpr scopes v) (f v)
-- To get respoce can it optimize for `v` first we are required to ask:
-- Can the function be optimized for previous argument, `f`?
-- ...
-- Looks like nope, `v` is used before `f`
```
```haskell
-- Well, lets look again, can it be optimized for `f`?
addStackFrames f v =

   (`withFrameInfo` f v) $ (`EvaluatingExpr` v) scopes
-- Yes it does, `f` gets consumed before `v`, and so create reusable `withFrameInfo f $ ...`.
-- Can it be optimized for `v` now?
-- Yes it can - `v` is a last argument and consumed as a pass-through twice, and in the very tail its type info gets dissected and processed.
```

Actually, I need to learn to use and read `core` to explicitly see these.

---

There were some studies done on random argument place reuse, but the complexity to solve that question is immense, especially comparing to the classical way which gives organization to everyone, is already studied, implemented, well developed and used everywhere is Haskell, and which optimization is much more direct and simpler to achieve.

It all derives from the proper argument order that simplifies the reuse optimizations detection to the compiler.

Most frequently Haskell maps function over the structure, the task is to make for the conveyor the most effective lambda term "module", the function is a lambda that runs some actions `fs` [with base case `cs`] on some content `xs`. When the `fs` and `cs` are reused between each map to new content piece - that of course allows GHC to reuse them as binary conveyors for the `xs`. And if the function does self-recursion - if the prefix of the list of arguments matches between recursions - the compiler simplifies the recursion into 1 assembly JMP instruction, which we call tail recursion.

This is what I basically do here last month. This is what function implementation changes are about, factoring-out of HOFs, function flips, and the code style.


But to be sure - I need to learn to use and read `core` to explicitly see these.

Anton-Latukha: ---

Of course relying on the profiler, because its results are more reproducible under the variable load on the running desktop system, and more directly comparable, than the benchmarks.

Run the check again, and reproduced the result.

Before change:

Screenshot-2021-03-10-19:48:41

After the simple optimization sectioning:

Screenshot-2021-03-10-19:49:22

Just as in the previous message seeing just that, was 2.4%, became ~1.6-1.7%, it is with profiler ~55% overhead.

As we see, results indeed reproduce and we have saved 15 seconds of execution with simple optimization.

Anton-Latukha: `cabal exec ghc-core -- -- -dsuppress-all -ddump-simpl ./src/Nix/Eval.hs`

In `core` language `addStackFrames` was:
```haskell
-- RHS size: {terms: 45, types: 109, coercions: 0, joins: 0/6}
addStackFrames
  = \ @ v_agpS
      @ e_agpT
      @ m_agpU
      @ a_agpV
      $dScoped_agpX
      $d(%,,%)_agpY
      $dTypeable_agpZ
      $dTypeable1_agq0 ->
      let { lvl28_smcX = currentScopes $dScoped_agpX } in
      let { $dHas_slML = $p2(%,,%) $d(%,,%)_agpY } in
      let {
        $dException_slMN
          = $fExceptionEvalFrame $dTypeable1_agq0 $dTypeable_agpZ } in
      let { $dMonadReader_slMK = $p1(%,,%) $d(%,,%)_agpY } in
      let { $dMonad_slMJ = $p1MonadReader $dMonadReader_slMK } in
      \ f2_acmQ v1_acmR ->
        let { lvl29_smcY = f2_acmQ v1_acmR } in
        >>=
          $dMonad_slMJ
          lvl28_smcX
          (\ scopes_acmS ->
             case $dMonadReader_slMK of
             { C:MonadReader ww5_ahAM ww6_ahAN ww7_ahAO ww8_ahAP ->
             $wwithFrame
               ww7_ahAO
               $dHas_slML
               $dException_slMN
               Info
               (EvaluatingExpr scopes_acmS v1_acmR)
               lvl29_smcY
             })
```

Became:
```haskell
-- RHS size: {terms: 32, types: 76, coercions: 0, joins: 0/2}
$waddStackFrames
  = \ @ v_sne2
      @ e_sne3
      @ m_sne4
      @ a_sne5
      w_sne6
      ww_snel
      ww1_sneq
      ww2_snet
      w1_sne8
      w2_sne9
      w3_snea
      w4_sneb ->
      let { lvl28_sktK = w3_snea w4_sneb } in
      let { $dException_sk47 = $fExceptionEvalFrame w2_sne9 w1_sne8 } in
      ww_snel
        (currentScopes w_sne6)
        (\ scopes_acoQ ->
           $wwithFrame
             ww1_sneq
             ww2_snet
             $dException_sk47
             Info
             (EvaluatingExpr scopes_acoQ w4_sneb)
             lvl28_sktK)

-- RHS size: {terms: 28, types: 98, coercions: 0, joins: 0/0}
addStackFrames
  = \ @ v_sne2
      @ e_sne3
      @ m_sne4
      @ a_sne5
      w_sne6
      w1_sne7
      w2_sne8
      w3_sne9
      w4_snea
      w5_sneb ->
      case w1_sne7 of { (ww1_snee, ww2_snet, ww3_sneu) ->
      case ww1_snee of
      { C:MonadReader ww5_sneh ww6_snep ww7_sneq ww8_sner ->
      case ww5_sneh of
      { C:Monad ww10_snek ww11_snel ww12_snem ww13_snen ->
      $waddStackFrames
        w_sne6 ww11_snel ww7_sneq ww2_snet w2_sne8 w3_sne9 w4_snea w5_sneb
      }
      }
      }
```

Where `$w*` - is a where statement.

layus: Nice and beautiful explanation, thanks ! I am quite surprised that GHC does not optimize that by itself, but your explanation makes sense. Nice catch :+1:

Anton-Latukha: And I'm not surprised to GHC, simply trying to help him, the `withFrame Info (EvaluatingExpr scopes v) (f v)` has reuse if 2-d & 3-d arguments are flipped, to solve it generally - is to solve a combinatorial problem and solution would do some sort of heavy multipass.

expipiplus1:
hahah, wtf is this: 


```diff
 annotateLocation :: Parser a -> Parser (Ann SrcSpan a)
-annotateLocation p = do
-  begin <- getSourcePos
-  res   <- p
-  end   <- get -- The state set before the last whitespace
-  pure $ Ann (SrcSpan begin end) res
+annotateLocation p =
+  do
+    begin <- getSourcePos
+    end   <- get -- The state set before the last whitespace
+
+    Ann (SrcSpan begin end) <$> p
```

This completely breaks location annotations! Guess that's my fault for not having any tests for them!

Anton-Latukha:
The continuation is in #936.

Indeed was guilty of it.

expipiplus1:
Thanks for releasing so promptly!

Issue #879, opened 2021-03-11
Please, apply YAGNI to the own implementation


Anton-Latukha:
First, lets say that implementing the project was very difficult task, and all my ramblings are not critique and not any detriment to the work done. I just discuss cleaning-up things after the creation stage. As my petty experience says - refactoring really in real world practice indeed takes 50% of the time of the coding. Frequently I make things work in 2.5 days, and then I clean-it-up 2.5 days until I am happy with the implementation.

Lets joke a little about our current state:

<details><summary> picture </summary>
</p>

![](https://www.monkeyuser.com/assets/images/2019/123-yagni.png)

</details>

It currently depicts the HNix engine implementation to what CLI currently does.

And the engine does not have a drag, it almost turns the blades.

The more work on the code polish done, the more YAGNI in the engine reveals themselves. The faster the engine turns, so - maybe we would be able to build a whole-lot-of-a-big-deal-plain on it that helps a lot of people.

---

The YAGNI needs to be reduced from the main computational paths and the main core of the engine, and in many cases, to save what additional functionality is already provided - we can easily preserve the current YAGNIs as such, optional, as in https://github.com/haskell-nix/hnix/issues/864.

Good example of them, is Kleisli actions that were drugged around in functions when the function of those functions was completely other thing. 

The https://github.com/haskell-nix/hnix/issues/864 and https://github.com/haskell-nix/hnix/issues/850 results are gargantuan, code inside the functions and around them became elegant, easy to use and understand, and moreover - more efficient, and allowed and allows more and more elegance.

Or in `Pretty` printing the outside package module function was used to place in its place a literal as `'.'` or `' '`, so with those function stacks the implementation and visual representation of the result were not understandable.

After the literals became literal - the code became much more straightforward to understand and work with.

---

For example: the `queryM` initially was implemented as:
```haskell
queryM :: MonadVar m => NThunkF m v -> m a -> (v -> m a) -> m a
queryM (Thunk _ active ref) n k = do
  nowActive <- atomicModifyVar active (True, )
  if nowActive
    then n
    else do
      eres <- readVar ref
      res  <- case eres of
        Computed v -> k v
        _          -> n
      _ <- atomicModifyVar active (False, )
      return res
```

There is a lot of mystery around it. `querryM` is quite a mysterious function. Was for me for the last 2 years of learning the project. I looked at it a couple of times and was not sure what its purpose is.
Its type signature and how to use it, why it is so and what this function details of the purpuse are a mystery.

[...]

PR #884, opened 2021-03-12
Refactors

Anton-Latukha:
Reducing some current YAGNIs.

Starting reducing some superfluous things in the code (https://github.com/haskell-nix/hnix/issues/879)

* `Result` type is really an euphemism to `Either`. `Left` `Right` is on the instinctive understanding for Hakellers, and using it enables further readability and plasticity to the code. 
Currently made `Result`, what it is - an euphemism to `Either`. Probably we may even just remove the custom type without any detriment, it probably would just make code even easier to read, since the `Either` semantics already fully embody what `Result` is, and the habitat of the function that brings `Either` shows where it came from. Outside of the module that exposes the data type `Result` is literally does not appear not once in type signatures anywhere over the project.

* `forceEff` changed a little, no reason to lock and then do preparation actions that do not required locking the whole process. So now doing preparing first and then locking right when it is needed.

* Please note that `force{,Eff}` implementations internally are now the same.

* `queryM` became non-locking, since read from static structures in a static language does not require locking operations.

---

Update:

* notice the implementation transformation for `further` - now it really shows the nature that all there was - it really simply defers the thunk. The question here is: since locking was not implemented for it, does `atomicModifyVar` needs thunk locking, or its atomicity allows this. But anyway, `further` family of functions used just 1-nce in a `inform` type cass instance that is not used at all.

* we arrived at that `force{,Eff}` are exactly 1 function, they are united currently under `forceMain`, the future (a cascade other MonadThunk instances above, for example, and future in general) would show does `class MonadThunk` needs `forceEff`, or `forceEff` is just `force`.


Anton-Latukha:
I have an idea.

src/Nix/Options/Parser.hs
```diff
@@ -62,7 +62,7 @@ nixOptions current =
         )
     <*> switch
         (  long "values"
-        <> help "Enable reporting of value provenance in error messages"
+        <> help "Enable reporting of value provenance in fail messages"
```
Ericson2314: unintentional replace?

Anton-Latukha: Thank you. This one is indeed.

Anton-Latukha:
Well.

Wanted to do something like:

Which would allow `force = operateOnThunk . queryM`, and implement locking as `force = (withLock operateOnThunk) . queryM`

But `readVar ref` requires MonadTrans.

```haskell
-- (roughly)
  -- | Non-blocking query
  queryM :: NThunkF m v -> ExceptT (m v) (IdentityT) v
  queryM (Thunk _ _ ref) =
    do
      (\case
        Computed v -> lift $ pure v
        Deferred a -> throwE a
        ) =<< readVar ref
```

So far not married the types.

And because of `MonadTrans` requirement, the result would be hardly faster or better than the current code.

I am still uncomfortable with `MonadTrans`, though I can construct and operate on them slowly, but the added additional hassle, code noisiness & complexity, etc. they do not smile to me. So I'm being lazy with remembering them keep forgetting them and as lazy re-learning them or touch them.

I know that what uses State monad with enough (sometimes big) persistence - that the data processing code can be written properly elegantly without it, but short-circuiting the I to O response to help debug the process may take some time.

There is a continuation-passing style.

I just do not love MonadTrans for so many reasons, being so very lazy with it.

Anton-Latukha:
In fact, talking about the continuation-passing style, instances I was refactoring were created with it.

So after creation, they were needed to be refactored inside-out to achive optimization.

Issue #555, opened 2020-05-19
Structure the code base with outshine (haddock struct) mode comments


Anton-Latukha:
`outshine` and `haddock` structure blocks are the same thing.
`outshine` : https://github.com/alphapapa/outshine is current upstream
Outstanding blog post: https://www.modernemacs.com/post/outline-ivy/
Worg: https://orgmode.org/worg/org-tutorials/org-outside-org.html

Example: my Emacs `init.el` outline tree:
![Screenshot_20200519_220443-1](https://user-images.githubusercontent.com/20933385/82367744-f3e73880-9a1c-11ea-9d5d-9f69066858f4.png)

... so in several seconds - now you know the major part of what is in my Emacs config code base, and you roughly have a feel of the logical structure of the code.

It allows to achieve in one action:
  * commenting
  * structuring
  * outlining
  * easier to move arround in the code
  * and so easier to work with
  * also by hiding unnecessary code and concentration on the parts at hand - increases direction and concentration
  * allows to understand the codebase in seconds
  * allows to understand the codebase better and its structure
  * It allows structural editing by `org-mode` functions
  * by that it enables also easy structural editing / reorganization the source code
  * people who do not have or not use outshine would receive comments of the structure of the code - so they also straightly win from it.
  * by all that - it is more pleasent to code now
  * all this accelerates contributions, newcomers learn, contribute faster and better - because they have an understanding now.

And that is achieved just by commenting the blocks of the code.

I am learning how packaging currently done in the project, and `dafault.nix` craves for the `outshine`.

It is a win - win - win, `more win` practice. It is literate programming in `*.hs`, even more:
```
∀ Languages, ∀ Texts : ∃ {Commenting system, Outlining, Structuring, Literate programming}
```

It brings only benefits.

So I just straight would open PRs for it and link them here.

Anton-Latukha:
In `nix-mode` and other modes where it does not work by layering comment symbols - the `outshine` works through the:

![Screenshot_20200519_224604](https://user-images.githubusercontent.com/20933385/82371324-abcb1480-9a22-11ea-9a51-19d7fb2e8160.png)

Anton-Latukha:
Oh yes. Nix has a lot of `{..};` , so it interhal headlines inside it - the closing `}` should be in separate heading - I would use `# **** <-here is the space char`, so during structural any rearengements of the blocks - `}` is not grabbed without the knowledge of the person.

[...]

PR #886, opened 2021-03-14
Harmonize NValue and NValue' notations, add doc ad COMPLETE annotations to patterns.

layus:
This should probably be considered as a draft.

I was investigating out NValue representation. It turns out that an NValue is a (Free NValue') but that NValue' constructor is NValue. Went around to turn the NValue' constructor to be NValue', and it turns there is only a few places where the change has an impact.

I also added some hands-on documentation on the tricky fiddling we do in there, with many type variables ultimately representing the same thing, but not at declaration time.

I would welcome contributions or pointers to proper haskell in-code documentation.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -137,9 +137,11 @@
         MonadPaths       (Fix1 t)    :: Nix.Standard -> Nix.Effects
         MonadPutStr      (Fix1 t)    :: Nix.Standard -> Nix.Effects
         ```
-  * [(link)](https://github.com/haskell-nix/hnix/pull/878/files) `nvSet{,',P}`: got unflipped, now accept source position argument before the value.
+  * [(link)](https://github.com/haskell-nix/hnix/pull/878/files) `Nix.Value`: `nvSet{,',P}`: got unflipped, now accept source position argument before the value.
   
-  * [(link)](https://github.com/haskell-nix/hnix/pull/878/files) `mkNixDoc`: got unflipped.
+  * [(link)](https://github.com/haskell-nix/hnix/pull/878/files) `Nix.Pretty`: `mkNixDoc`: got unflipped.
+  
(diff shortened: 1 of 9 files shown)


Anton-Latukha:
Nice, indeed. There was some puzzlement there before. Now we see it is `NValue'` there.

+====

About documentation:
The best live examples for documentation for the project that know of is https://github.com/kowainik/relude, there complete documentation and then more. Over `kowainik` projects overall they keep extremely good documentation, so they ironed-out how to use the majority of the Haddock features and layouts. Those sources are a good example to look into for some particularities of the practice.

One thing that I would differ with them, they section the imports, while I prefer sectioning the source code, since then people have structural awareness of where they are and what the main theme of the code block while reading the code and working in the source files. And that simple source outlining during work gives organization overall, a peace of mind, structural folding, editing: 

<details><summary>Screenshot</summary>

Screenshot-2021-03-15-14:16:45

It was #555 in the past.

</details>

About Haddocks, of course, the practice makes perfect. Myself still learning the specific details of Haddock.

Also, you've see probably some of my `Value` docs, I knew they were corny at the time of writing, but they came out too corny :rofl:.

If you consider the patch a draft, then lock it in as a draft. Making the patch ready for merge and right-away stating in the head post that it is a draft - sends opposite messages. Why it is a draft, I see it complete. Probably you consider it as a draft, since docs can be improved.

I began to see, like the Kernel team says that security bugs are just regular bugs, that the documentation process is just code refactoring, meaning that it is sneaked-in duringx the main work. Refactors to docs are even lighter questions to the refactors to code and they both a kinda moral, sanitary, and effective necessity - the most effective way to go about them is to do the docs and code refactors as drive-by improvements into the patches during main work.

For me it looks good to merge. As personally with docs I adhere to the Wiki community post-editing and the rule: having documentation is better than not having it. And polishing the docs in the environment of merge requests I consider it to be overall a very counter-productive and socially awkward process and awkward overall. Rather then doing that, I like saying "thank you" and merging the improvements to docs.

Saying "thanks" and eagerly merging docs also goes into approaching to docs as code refactoring.

Even go further - saying that if a person here was promoted to be a maintainers, that person is qualified enough to give them *carte blanche* to self-merge docs. In that way, the docs processes and quality would ramp up greatly. Up until the world would start notifying us and shown that the changes to that are required.

PR #899, opened 2021-03-23
Builtins refactor: normalizing the builtin fun names

Anton-Latukha:
Noticed that the `*_` was used to avoid collision with well-known Haskell
function names.

But `*_` has a different semantic meaning in Haskell, and the reader still thought it
is Haskell well-known functions with `_` modification.

This also created a dichotomy in function names for the `builtins`, now some were normal, and some
were `_` without difference in the semantic meaning to them.
So reader logic was tripping over this dichotomy and between the well-known Haskell names and their `_` modifications.

`builtins` functions are internal to module. Now they are formed in `*Nix` subgroup.

This subgroup can be put in a separate module and imported qualified, and so -
would be able to drop the `*Nix` suffix from them.

### DIFF
--- a/src/Nix/Builtins.hs
+++ b/src/Nix/Builtins.hs
@@ -160,108 +160,108 @@ builtinsList = sequence
       version <- toValue (5 :: Int)
       pure $ Builtin Normal ("langVersion", version)
 
-  , add  TopLevel "abort"            throw_ -- for now
-  , add2 Normal   "add"              add_
-  , add2 Normal   "addErrorContext"  addErrorContext
-  , add2 Normal   "all"              all_
-  , add2 Normal   "any"              any_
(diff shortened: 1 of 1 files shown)


layus:
Nice one. That being said, I would have expected the names to be nix* rather than *Nix.
Not really important though ;-)

Anton-Latukha:
Well, ^Nix was easier to do. & suffix use for me seemed more suggestive mentally with direct correspondence to Haskell functions. Like: "this is the same things, but in Nix context".

---

Since it is the biggest module, wanted to split-away the functions into `Builtins.Function` module. But nope. `builtins.builtins.builtins...` is a part of the code design. Left a note that the module not really can be split that nicely.

PR #903, opened 2021-03-23
Use a read-only store

layus:
This is an example of how we could make hnix use a read-only store because it makes little sense to use the remote daemon in the current state of the project.

1. Using the daemon makes tests complex
2. We cannot do anything useful with the daemon atm (except generating raw files and .drv in the store)
3. The daemon is not used as intended, as we do not persist the connection across calls.

See https://github.com/haskell-nix/hnix-store/pull/144

### DIFF
--- a/src/Nix/Effects.hs
+++ b/src/Nix/Effects.hs
@@ -44,8 +44,8 @@ import qualified System.Info
 import           System.Process
 
 import qualified System.Nix.Hash               as Store
-import qualified System.Nix.Store.Remote       as Store.Remote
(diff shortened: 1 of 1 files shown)


sorki:
Good idea! Daemon tests can wait until I manage to decouple `-remote` test framework and make it usable by `hnix` as well.

layus:
Got stale, but the rebase is fairly trivial. Still waiting on a release of hnix-store-core >= 0.4.3 in nixpkgs in order to get this merged.

PR #906, opened 2021-03-25
treewide: light relude linting

Anton-Latukha:
Provided `relude's` `.hlint.yaml` locally.

To my taste most of the suggestions are specific to `relude` and can be
considered harmful for future portability.

Only the `(fmap . fmap) == <<$>>` & `*.singleton == one`, I'd considered
somewhat reasonable to maybe thinnk about adopting.

The flip of the arguments in `whenJust` and so on - surprised me.

These here are what indeed is light and helps to clean-up a bit and in future
text data types & list migrations.

Most of the `relude` HLint rules I saw, I want to remove from the checks. So decisively not going to implement them
in CI.


Anton-Latukha:
Ok, the infra got a light fix for the time being.

PR #908, opened 2021-03-27
to Text migrations; NixString code refactors

Anton-Latukha:
Doing `String -> Text` conversions, but got into optimization weeds.

`NixString` is a type where content and context always do everything and change together, it is a best-case of rarely encountered tuple laziness & recursion. So couldn't walk past it.


Anton-Latukha:
Like the first time in forever seem to hit the definitive internal difference between GHC versions. [GHC 8.6](https://github.com/haskell-nix/hnix/pull/908/checks?check_run_id=2218506189) when GHC 8.10 infers normally there.

Interesting.

Indeed, the most probable difference to observe is when with TH to reach deeper into the depth of the compiler, or probably some low-level with unsafeness and levity polymorphism, stuff I currently know nothing about.

Anton-Latukha:
Well, actually it was just a missing instance.

It is interesting and useful to note that Lift Text is supplied by default from GHC 8.8.

Anton-Latukha:
It is pretty much all easily replaceable String -> Text.

Did not made the FilePath -> Text. So far. Want to approach them in a bulk.

The String cases left need more detailed attention.

There is also left the zone in Effects, because there is a lot of Strings there, and some of them are in classes and such, intertwined also, probably I simply got tired in the evening and PR got long, just want to merge these, and work further on zones mentioned above.

Issue #912, opened 2021-03-31
Please, migrate FilePaths (String) to Path (from path)


Anton-Latukha:
During `(String -> Text)` migration, inevitably the question is what to do with the `FilePath` arose.

What I've thought to do, or to have `type PathText = Text`, or to have an equivalent situation for it, but in Text.

Of course the POSIX libs & base below are String anyway [the case](http://blog.ezyang.com/2016/09/the-base-of-a-string-theory-for-haskell/).

The author of the `path` package and `Path` data type is Snoyman, who with *hvr* & Neil Mitchel are coauthors of the [Abstract FilePath Proposal (AFPP)](https://gitlab.haskell.org/ghc/ghc/-/wikis/proposal/abstract-file-path).

Overall the `path` is one of the currently frequently used packages.

So in the face of a new base being done, it seems like a good strategy to start doing work now to migrate to `FilePath` to `Path`, because we would receive exactly `Path` or very close equivalent to it in the new `base`.

Since currently the:
```haskell
newtype Path b t = Path FilePath
```

Both the data type logic should respect it as a path, and casting to Text is possible.

Because currently in the code in some places the boundary between FilePath and String does not exist, some functions that are to accept only FilePath call it String.

Migration to `newtype Path b t = Path FilePath` should make logic more robust and establish the proper type and type casting boundary that would allow further easy migrations for ways the path data types would go. And casting it into text is just as possible.

Currently I just want to finish the `(String -> Text)` and would go around the `FilePath` for the time.

Anton-Latukha:
Ok.

I've just done the pass over the source code.

From what is left after migration, the most internal `String` and `toText _` is `FilePath`. And there was no type boundary between `FilePath` and `String` and after I `(String -> Text)` - where the FilePath becomes textual type became even vaguer.

It became obvious that `FilePath` should be casted to some type that makes the boundary real, so the path handling functions become specified to the path handling, and textual handling to such, and so the safe type casting happens in between.

`path` also provides functionality to distinguish between path types, which HNix currently does internally.

Overall, this is the next move in the transition from String in the HNix.

sternenseemann:
Note that this still has the somewhat problematic assumption that all Paths are valid Unicode in some capacity. This is not correct for POSIX as paths are arbitrary byte sequences that don't require any kind of encoding. This would be an issue with nix in some (probably) corner cases since Nix strings are also just arbitrary byte sequences without any requirements in terms of encoding.

[...]

PR #915, opened 2021-04-03
toText refactors

Anton-Latukha:
* What is currently left to migrate is the last realistic bit currently, the more complex String manipulations that initially passed by, and now would return and migrate it.

* What then what is left is Store Strings, FilePaths, and places that have heavy external String use, it is pity that a lot of libs give only Strings.
* Let's hope that more updates would arrive with current endeavors.
* We collectively as Haskellers need to pick upstreams to migrate to modern text types.

After the mentioned at the top migrations, would consider the migration as largely done in HNix, and would move to other things, probably to do what is possible for the GHC support.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -41,6 +41,7 @@ import           Prettyprinter.Render.Text
 import qualified Repl
 import           System.FilePath
 import qualified Text.Show.Pretty              as PS
+import           Nix.Utils.Fix1                 ( Fix1T )
 
 main :: IO ()
 main =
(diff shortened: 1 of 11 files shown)


Anton-Latukha:
Over the "to Text" migrations run rough real-life `Firefox` evaluation "benchmarking", the time of computation was roughly the same across the points in migration.

Thou looked into time performance, since was more worried about maybe some marshaling somewhere, but currently, as put above - seems not. There should be quite a memory save, Text is much more memory efficient and strict type at the same time.

During these last overpasses, the #912 is the main thing and became apparent.

PR #918, opened 2021-04-30
deps: ref-ft upd, Fresh: rm instance MonadAtomicRef

Anton-Latukha:
Upstreamed instance update arrived. `ref-tf` has unrestrictive bounds, so backward compatible, so just removing the instance.


Anton-Latukha:
Too fast. Not in Nixpkgs at the moment. Really soon should be.

Anton-Latukha:
Ok.

`ref-tf` is in Stackage. Already in Stackage Nightly.
Nixpkgs currently on the version of the last week. Actually, it is hard to say the order, since Stackage Nightly & Nixpkgs update arrived simultaneously.

So waiting for roughly a week.

PR #936, opened 2021-05-22
Fix source locations

expipiplus1:
Add small test for source locations

### DIFF
--- a/src/Nix/Parser.hs
+++ b/src/Nix/Parser.hs
@@ -609,9 +609,10 @@ annotateLocation :: Parser a -> Parser (Ann SrcSpan a)
 annotateLocation p =
   do
     begin <- getSourcePos
+    res <- p
     end   <- get -- The state set before the last whitespace
 
-    Ann (SrcSpan begin end) <$> p
+    pure $ Ann (SrcSpan begin end) res
 
 annotateLocation1 :: Parser (NExprF NExprLoc) -> Parser NExprLoc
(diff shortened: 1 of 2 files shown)


expipiplus1:
Would be great to have a patch release with this. @Anton-Latukha :)

Anton-Latukha:
Great.

<details><summary>Tangent:</summary>

</p>
I am interested to get more understanding of source positions in interpreters&compilers.
From the code refactoring, I am under impression that for example order `NAttrSet (AttrSet SourcePos) (AttrSet (m v))` is more efficient since position changes slower than value. For example forcing a thunk - changes the `AttrSet (m v)`, but not `AttrSet SourcePos`.

---

</details>

> Would be great to have a patch release with this. @Anton-Latukha :)

Would release it in a couple of hours.

expipiplus1:
Thanks!

Anton-Latukha:
Well, bummer, rechecked, put the wrong comment during squash.

Then since I remember that line, rechecked, 

I remember it, I was doing the monad law across the project, and that one case was not one. Guilty, left it dangling while was doing other stuff (thought review the place), forgot that put it on the back burner and at some point it got in with the other change.

So that is my fail.

Of course I do not pretend that are none others, with this volume - there should be. As some point want to go into property writing/testing, since it is directly connected to the language/interpreting design.

Anton-Latukha:
Fix release is published: https://hackage.haskell.org/package/hnix-0.13.1

Thank you.

PR #939, opened 2021-05-23
Correct lower bound on semialign

expipiplus1:
before 1.2 it didn't expose the Data.Semialign.Indexed module

### DIFF
--- a/hnix.cabal
+++ b/hnix.cabal
@@ -443,7 +443,7 @@ library
     , regex-tdfa >= 1.2.3 && < 1.4
     , relude >= 1.0.0 && < 1.1.0
     , scientific >= 0.3.6 && < 0.4
-    , semialign >= 1 && < 1.3
+    , semialign >= 1.2 && < 1.3
     , serialise >= 0.2.1 && < 0.3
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
#### Automation of this is currently not possible

Overall it is strange to manage low bounds, especially since there is really no way to check them but manually. Cabal so far does not have any options (as far as known to my module) to prefer low versions. If it was - then it would've been possible to check the low bounds.

+====

#### Tangent
(*talking about [a worm place](https://www.youtube.com/watch?v=pw3PQEw6O4U) <sub>(the dependency hell)</sub>: imagining managing all these situations when HNix was also on Stackage which hardcode-influences the situation in Nixpkgs, some of these requirements would've been impossible to manage, the only way would've been to enter Haskell ecosystem technical debt*)

Now at least we concentrate on Hackage, can do GHC 9.0 work, and regularly check Nixpkgs - and that is pretty much it.

+====

#### Made according revisions

Made a revisions to the last releases (`0.13.1`, `0.13.0.1` && left `0.12.0.1` in the old state, since `semialign-indexed` is there).

expipiplus1:
Thanks!

PR #954, opened 2021-06-14
Main.hs: refactors

layus:
This is an opinionated refactor. First to get rid of `opts` interspersing, and also to try a new way to write cascading maybe. It keeps vertical align between equal expressions instead of the runaway indent of nested expressions that is all too common in haskell, and disrupts the important patterns.

Uses ad-hoc operators abundantly. Do you like it ?

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -2,8 +2,9 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE RecordWildCards #-}
 
-module Main where
+module Main ( main ) where
 
 import           Nix.Utils
 import           Control.Comonad                ( extract )
@@ -46,59 +47,61 @@ main =
     time <- getCurrentTime
     opts <- execParser $ nixOptionsInfo time
 
-    runWithBasicEffectsIO opts $ execContentsFilesOrRepl opts
+    main' opts
 
+main' :: Options -> IO ()
+main' (opts@Options{..}) = runWithBasicEffectsIO opts execContentsFilesOrRepl
  where
-  execContentsFilesOrRepl opts =
-    maybe
-      (maybe
-        (maybe
-          (list
-            (withNixContext mempty Repl.main) -- run REPL
-            (\case
-              ["-"] -> handleResult opts mempty . parseNixTextLoc =<< liftIO Text.getContents
-              _paths -> traverse_ (processFile opts) _paths
-            )
-            (filePaths opts)
-          )
-          (\ x ->
-            -- We can start use Text as in the base case, requires changing FilePath -> Text
-            traverse_ (processFile opts) . String.lines =<< liftIO
-              (case x of
-                "-" ->  getContents  -- get user input
-                _path -> readFile _path
-              )
-          )
-          (fromFile opts)
-        )
-        (handleResult opts mempty . parseNixTextLoc)
-        (expression opts)
-      )
-      (\ path ->
-        do
+  execContentsFilesOrRepl =
+    firstJust
+      -- The `--read` option: load expression from a serialized file.
+      [ readFrom <&> \path -> do
           let file = addExtension (dropExtension path) "nixc"
-          process opts (pure file) =<< liftIO (readCache path)
-      )
-      (readFrom opts)
+          process (Just file) =<< liftIO (readCache path)
 
-  processFile opts path =
-    do
-      eres <- parseNixFileLoc path
-      handleResult opts (pure path) eres
+      -- The `--expr` option: read expression from the argument string
+      , expression <&> processText
 
-  handleResult opts mpath =
+      -- The `--file` argument: read expressions from the files listed in the argument file
+      , fromFile <&> \x ->
+          -- We can start use Text as in the base case, requires changing FilePath -> Text
+          traverse_ processFile . String.lines =<< liftIO
+            (case x of
+              "-" -> getContents
+              fp -> readFile fp
+            )
+      ]
+    `orElse`
+      -- The base case: read expressions from the files listed on the command line
+      case filePaths of
+        -- With no files, fall back to running the REPL
+        [] -> withNixContext mempty Repl.main
+        ["-"] -> processText =<< liftIO Text.getContents
+        _paths -> traverse_ processFile _paths
+
+  firstJust :: [Maybe a] -> Maybe a
+  firstJust = asum
+
+  orElse :: Maybe a -> a -> a
+  orElse = flip fromMaybe
+
+  processText text = handleResult Nothing     $   parseNixTextLoc text
+
+  processFile path = handleResult (Just path) =<< parseNixFileLoc path
+
+  handleResult mpath =
     either
       (\ err ->
         bool
           errorWithoutStackTrace
           (liftIO . hPutStrLn stderr)
-          (ignoreErrors opts)
+          ignoreErrors
           $ "Parse failed: " <> show err
       )
 
       (\ expr ->
         do
-          when (check opts) $
+          when check $
             do
               expr' <- liftIO (reduceExpr mpath expr)
               either
@@ -111,7 +114,7 @@ main =
                 -- liftIO $ putStrLn $ runST $
                 --     runLintM opts . renderSymbolic =<< lint opts expr
 
-          catch (process opts mpath expr) $
+          catch (process mpath expr) $
             \case
               NixException frames ->
                 errorWithoutStackTrace . show =<<
@@ -120,7 +123,7 @@ main =
                     @(StdThunk (StandardT (StdIdT IO)))
                     frames
 
-          when (repl opts) $
+          when repl $
             withNixContext mempty $
               bool
                 Repl.main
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
Nested maybes still read pretty much the same whey the nested cases were.
But because Lisp-slype is easy to be reasoned about to reorganize & maybe convert into `do`, also edit/refactor even can be template/macrosed.

Nested layering can be observed and swapped around selecting a blocks of code, & maybe [flattening code with leveraging a `do` wrapping](https://www.haskellforall.com/2021/05/the-trick-to-avoid-deeply-nested-error.html) or use one-handed control flow operators as `when` & [these](https://hackage.haskell.org/package/extra-1.7.9/docs/Control-Monad-Extra.html#v:whenJust) or [these](https://hackage.haskell.org/package/cond-0.4.1.1/docs/Control-Conditional.html#v:when) or Arrows, in other words this style allows to see and find useful control flow combinators to adopt into the programming style.

<details><summary> Note on the use of reversed operators </summary>

</p> 
I'm definitely not a fan of reversed control flow operators, as `&`, `<&>`, `??`, `>>=`. Applying argument `x` as `x f` always feels weird to me, as Lambda Calculus, System F, Haskell Core & so Haskell language has diametrally opposite design.

I kinda see them as "it was convenience during creation" type of operators when for people is simpler to think in that way to circumvent the Haskell design. But it does a disservice to contributors afterward, lets answer "Why?".

So far very rarely I saw reversed control flow operators really stay & fit in & after refactoring. It is harder to reason & read through them anyway (which I touched on in the dialog). & often I find that refactoring them into `$`, `<$>`, `()`, `=<<` - aligns things & shows how to do simplification in refactoring. A lot of refactoring to conciseness done in HNix was due to me sitting & flipping control flow fighting operators into a normal position to align code flow to Haskell design & because code became aligned with the design of the language - further Haskell itself shows the way to refactor & how to modify & extend the code.

It was "fun" flipping several hundred of `>>=` into `=<<` - and that action sadly can not be automated through action macrosing. Just as it was (strangely enough) not possible to automate by macros a `>>=` into `do` notation.

But after several hundred of manual refactors of `>>=` into `=<<`  - in the normal position the conversion of `=<<` into `do` was easily automatable. & those `do` blocks allowed to leverage constant advanced support of optimization of future development through `ApplicativeDo`. Once it got wrapped into `do` - further after even the most complex developments in that code would be optimized by advanced compiler analysis automatically.

Just giving an example of how far the ball throws itself in Haskell if to use Haskell operators that align with Lambda calculus control/reading flow.
 
 </details>
 
 But despite my view on those operators - it is not that important.
 
As it is much more important than my one-sided work gets counter-argumented, as no man is an island. & and it is much more important that you provided improvements.

Local ad-hoc operators are a good practice.

layus:
Thanks for merging. I am undecided about this lisp style for as much as I understand the benefits in terms of rewriting (and ultimately execution speed*) I find it extremely difficult to read because the code gets a weird shape, related parts get separated and patterns are more difficult to spot.

As a not so bad example, look a line https://github.com/layus/hnix/blob/e1e1eff1e0da5402685450faa78a3ff9a2f90ece/main/Main.hs#L74 and try to guess when it is actually called, and with what argument. I find it quite difficult. I am getting used to it, so I do not care much either, but I wonder if we prefer a codebase that is easy to read, or easy to refactor.

At the moment it seems we need refactors over readability. We can always switch back later on :-)

Anton-Latukha:
> Thanks for merging. I am undecided about this lisp style for as much as I understand the benefits in terms of rewriting (and ultimately execution speed*) I find it extremely difficult to read because the code gets a weird shape, related parts get separated and patterns are more difficult to spot.

Well, it is right diametrally contrary for me. Do you read Haskell expressions from tails to head?

> As a not so bad example, look a line https://github.com/layus/hnix/blob/e1e1eff1e0da5402685450faa78a3ff9a2f90ece/main/Main.hs#L74 and try to guess when it is actually called

It all processes the `(readFrom opts)`, one needs to read what argument is processed, argument applies the end of expressions and then reads above.

This was just as nasty in case statements also. I can argue this piece is harder to read than it can be with the main reason being that CLI fields have no documentation & so it is hard to say what is a semantic difference between `readFrom :: Maybe FilePath` & `fromFile :: Maybe FilePath`.

Lisp-style code always can be greatly simplified by simple sectioning, as it easily be sectioned in:

```haskell
    let
      maybeReadFromGiven =
        maybe
          maybeExpressionGiven
          (\ path ->
            do
              let file = addExtension (dropExtension path) "nixc"
              process opts (pure file) =<< liftIO (readCache path)
          )
          $ readFrom opts

      maybeExpressionGiven =
        maybe
          maybeFromFileGiven
          (handleResult opts mempty . parseNixTextLoc)
          $ expression opts

      maybeFromFileGiven =
        maybe
          maybeListOfFilesGiven
          (\ x ->
            -- We can start use Text as in the base case, requires changing FilePath -> Text
            traverse_ (processFile opts) . String.lines =<< liftIO
              (case x of
                "-" ->  getContents  -- get user input
                _path -> readFile _path
              )
          )
          $ fromFile opts

      maybeListOfFilesGiven =
        list
          runRepl
          $ filePaths opts

      runRepl = withNixContext mempty Repl.main
```

And now suddenly it gets easy to work on it & refactor. I put `$` there to underline that the very tails of expressions (from which reading of expressions starts) itself now show that `opts` in fact can be applied externally to all those expressions. & we began to see.

Then it logically pushes us to think how to fold this. Those `maybe` just need to short-circuit on `Just`. Well, bummer - `Monad` short-circuits on `Nothing`. But, is there no other way to short-circuit...?, is Haskell is *so poor in its plasticity* there *really* just nothing can short circuit on `Just` & no one thought about it? We think a bit more & we remember there is something that short-circuits on `Just`:
```haskell
ghci> Nothing <|> Just "A"
Just "A"`
```
We replace lines of previously nested calls with `mempty` (or `Nothing` if needed) & now they are suddenly actually just a `<|>` chain!:
```
  maybeReadFromGiven <|> maybeExpressionGiven <|> maybeFromFileGiven <|> maybeListOfFilesGiven <|> runRepl
```

We made 1 obvious structural action - decomposition & Haskell right away shouts to us that it is an `Alternative` chain. We made a 2-nd structural action - use `mempty` & got an `Alternative` chain.

Profit!?..

& actually, we used to have `whenJust = maybe mempty` in `Utils`. Removed due to name clash with `relude` `relude: whenJust = maybe (pure ())`, I think they made an error there, as `maybe mempty` is a much more general & frequent pattern, as we see - it integrates with `Alternative`.

Should we take-out `opts` out of them? *idk*, `opts` is stored once in a main function global scope, & subfunctions just access different fields, for GHC it is trivial (pointer) to wire them up to it. Is it worth taking them out would depend on would we see other gains to it & the answer frequently is yes.

layus:
Sectioning looks like a good solution in this case. More names (which is a good and a bad thing altogether), and the code is not forced to stay in the same order as the choices we make, but it 
avoids the other pitfals.

`Just re-refactor :: Maybe WillDo` :-D

Anton-Latukha:
> At the moment it seems we need refactors over readability.

I recommend reading a head of the function (well, due to Lambda Calculus binding design) & then read from input to output - meaning from tail where arguments come in, to the head, where the last actions are done on them.

<details><summary> Lambda Calculus design bug </summary>

</p>
That Lambda Calculus function bindings arguments in the head & apply them in the tail, I think is a design issue of Lambda Calculus functions.

As:
```haskell
  ($)
    =  f a
    <- f a
  ($)
    :: b
    <- (b <- a)
    <- a
 
  (.)
    =  g f
    <- g f
  (.)
    :: c
    <- (c <- b)
    <- (b <- a)
  
  function
    (someLong
      function
        f
        $ somewhereHereItTakes
          a
    )
    = f a
  function
    :: b
    <- (b <- a)
    <- a
    
  function2
    (f <$> a)
    = f a
  function2
    :: b
    <- (b <- a)
    <- a
    <= Functor a

 -- ^ Suddenly aligns everything, but it reads in reverse, due to mathematicians chosen to write `g . f x`
```

Is structured better & reads better, if people learned to read it that way. & also preserves all lambda function saturation optimizations Haskell have. & Haskell code essentially does not need changes except for type signature switcharoo & putting the lambda binding at the end (where it applies right away). & now functions to the reader the first thing tell what it produces, then, if reader still needs more info then the first word - shows what it actually does to produce result. And requirements/constraints get into the proper "where" section below the main type, just as we do in good scientific literature.

Also:

```
function
  g . f
  = g f
function
  :: (c <- b)
  <- (b <- a)
```

It is due to how function <- argument application goes & composition goes.

If we'd to flip the direction of functional application & composition we use - then it would be logical to reverse *this design.

But with functional application & composition from tail to the beginning - Lambda function binding is one being weird there.

People have difficulty with Haskell because of this Lambda Calculus notation design bug & some Haskell design decisions on top.



---

I also kinda thinking about to play/build such Lisp DSL one day, just to see if my reasoning is sound.

</details>

Anton-Latukha:
> Sectioning looks like a good solution in this case. More names (which is a good and a bad thing altogether), and the code is not forced to stay in the same order as the choices we make, but it
> avoids the other pitfals.
> 
> `Just re-refactor :: Maybe WillDo` :-D

No worries, I just ask people to try to use `=<<` more & to think that Haskell is essentially a very fancy Lisp & that in that there may be something there. I know my code style is weird to how it is accustomed to write Haskell code.

I have chosen to stick to the Lambda Calculus design, except that disagreed & disregarded its binding decision. Between Lambda function binding order & function application & composition, I decided that in Haskell order of composition & function application are more important than lambda bind to lambda body order. It is how one gets insured in Lisp-style. & diagonal layout allows showing structure & access to things easier.

Again would reiterate - there is Haskell <- Haskell Core <- Lisp/SystemF <- Lambda Calculus <- Function composition <- Function application <- and as we see they have chosen to go `<-`. Well, again, Haskell function types go `->`... but well... We accustomed to it.

& it is impossible to escape this language grammar while talking with a compiler that talks & thinks in those languages.

Lisp is the most flexible (& also for refactoring) language there is. Currently, the Lisp-style made its main purpose - almost the whole of HNix got refactored & in the state I am comfortable for its future. & style would take a test of time, maybe some people would see something in it & see its cornerstones & purpose.

Most of the parts of the code should be much easier to read than one we toughed upon (I kinda know that part). Now project code sticking to this style (to have massive clean-up/refactoring by one person) becomes not that important, so please, do as you see, that is why I did it in the first place.

& maybe we even can put an automatic code formatter on it.

But for me, what I obtained from all those refactors - is a code style, & I got insured (for myself) it is a right style (at least for me) to guide refactoring & would write in it generally, at least for myself.

PR #960, opened 2021-06-23
Fix some issues revealed by static analysis

layus:


layus:
I had a run with [stan](https://github.com/kowainik/stan) an I have uploaded the [stan.html](https://maudoux.be/stan_original.html) report.

The tool is a bit heavy on strict data and avoiding unnamed black holes in pattern matches. But at least it made me remove some uses of Unsafe.

Another useful but weird case is that it highlights issues with the list type which theoretically allows infinite lists. Functions like length may thus not terminate. In our case moving to slists is not a good idea, but Vectors may represent better the lists we handle. But it is not worth switching if we create vectors from lists.

Anton-Latukha:
Good changes.

Recently I became aware & interested in how much the space leaks are a pinnacle of Haskell programming. I refuted space leaks & was concentrating on performance profiling. Until I understood that main Haskell laziness downside is thunk suspension resulting in space (& performance) leaks. To the point that would resume writing pause to write about it.

Seems like to be a good Haskell hackers - we need to learn to analyze the main laziness problem.

A nice article on the theme: https://queue.acm.org/detail.cfm?id=2538488

Stan is also interesting to see, first time meeting it in use.

<details><summary> [] & its length </summary>

</p>
About `length` problem - it is a problem of `[]` data type implementation.

Couple of days ago, playing with recursive shemes & implementing Peano numbers & List as:
```haskell
data ListF a r
  = NilF
  | ConsF a r
  
type List a = Fix (ListF a)
```

This creates a proper list datatype for Haskell, it has proper lazy properties. Its `"length"` also is lazy. Haskellers talk about that `[]` is not a proper list for Haskell - but, ironically, they do not direct to proper implementation & there is no use of the proper implementation.

</details>

Anton-Latukha:
macos build repaired.

If the build problem would return - would try to move the build to solve it.

Anton-Latukha:
> unnamed black holes in pattern matches

well, they are a real problem.

PR #985, opened 2021-08-14
Merge Prelude & Nix.Utils

Anton-Latukha:
We essentially had 2 preludes, now they are merged & allow to control what parts of `relude` are exposed.
It clean-up imports quite a bit & provides a set of frequently used functions.

`relude` is a pretty good prelude, but it still has flaws (for example it exposes a mix of `Text` & `String` functions, which is weird from the outside). If we want to use `Text` internally - that should be it, conversion can still be needed (for `base` ^ GHC reasons), but at least is would be only on the IO margins (which are `prelude` functions), not 2 conversions in the middle of data flow.

This setup helps to program with supplying `Text` type functions & support of stronger types.

There may be a specific thing - currently, it is probably tricky to import our `Prelude` module into other projects, that can be addressed by moving stuff back into `Nix.Utils` & passing it through `Prelude` module, but then people would start importing `Nix.Utils` from time to time in the code base, which is Ok, since because of `Prelude` passthrough - that would be rare.

I needed to do this sort of merge, to align all code & project this way properly, before I do further things with it.

This is the last thing I had in mind, I plan to stop refactors any moment if #804 ever continues.


Anton-Latukha:
Now we have both the most flexible `Prelude` setup & the `Nix.Utils` that holds custom code, nice.

PR #987, opened 2021-10-11
clean-up & refactors; enabling some tests; specializing Path functions

Anton-Latukha:
Specialization of former `FilePath` functions to `Path` happens mainly to:
1. Prune String from code as much as possible and to maximally avoid conversions in the end.
2. To have typesafe code (I/somebody still should remove `Monoid` for `Path`).
3. Main reason: further to have an easy decision making how to control & process `Path`s and to have breather migration to `path` package (probably), which would make it even more realistic & typesafe code.

### DIFF
--- a/src/Nix/Parser.hs
+++ b/src/Nix/Parser.hs
@@ -359,7 +359,7 @@ identifier =
   identLetter x = isAlphanumeric x || x == '_' || x == '\'' || x == '-'
 
 nixSym :: Parser NExprLoc
-nixSym = annotateLocation $ mkSymF . coerce <$> identifier
+nixSym = annotateLocation $ mkSymF <$> coerce identifier
 
 
 -- ** ( ) parens
@@ -822,7 +822,7 @@ nixSelect term =
 -- ** _ - syntax hole
 
 nixSynHole :: Parser NExprLoc
-nixSynHole = annotateLocation $ mkSynHoleF . coerce <$> (char '^' *> identifier)
+nixSynHole = annotateLocation $ mkSynHoleF <$> coerce (char '^' *> identifier)
 
 
 -- ** Expr & its constituents (Language term, expr algebra)
--- a/src/Nix/Parser.hs
+++ b/src/Nix/Parser.hs
@@ -812,11 +812,10 @@ nixSelect term =
       , Maybe NExprLoc
       )
     -> NExprLoc
-  build t mexpr =
+  build t =
     maybe
       t
       (\ (a, m) -> (`annNSelect` t) m a)
-      mexpr
 
 
 -- ** _ - syntax hole
--- a/src/Nix/Parser.hs
+++ b/src/Nix/Parser.hs
@@ -249,13 +249,13 @@ nixUri =
         do
           start    <- letterChar
           protocol <-
-            takeWhileP Nothing $
+            takeWhileP mempty $
               \ x ->
                 isAlphanumeric x
                 || (`elem` ("+-." :: String)) x
           _       <- single ':'
           address <-
-            takeWhile1P Nothing $
+            takeWhile1P mempty $
                 \ x ->
                   isAlphanumeric x
                   || (`elem` ("%/?:@&=+$,-_.!~*'" :: String)) x
@@ -431,12 +431,12 @@ pathStr :: Parser Path
 pathStr =
   lexeme $ coerce . toString <$>
     liftA2 (<>)
-      (takeWhileP Nothing pathChar)
+      (takeWhileP mempty pathChar)
       (Text.concat <$>
         some
           (liftA2 Text.cons
             slash
-            (takeWhile1P Nothing pathChar)
+            (takeWhile1P mempty pathChar)
           )
       )
 
@@ -491,7 +491,6 @@ opWithLoc f op name =
   do
     AnnUnit ann _ <-
       annotateLocation1 $
-        {- dbg (toString name) $ -}
         operator name
 
     pure . f $ AnnUnit ann op
@@ -640,6 +639,7 @@ getBinaryOperator = detectPrecedence spec
     \case
       (NBinaryDef assoc op name, _) -> [(op, OperatorInfo i assoc name)]
       _                             -> mempty
+
 getSpecialOperator :: NSpecialOp -> OperatorInfo
 getSpecialOperator NSelectOp = OperatorInfo 1 NAssocLeft "."
 getSpecialOperator o         = detectPrecedence spec o
--- a/src/Nix/Parser.hs
+++ b/src/Nix/Parser.hs
@@ -862,8 +862,6 @@ nixExprAlgebra =
       nixOperators nixSelector
     )
 
--- | Nix term parser.
--- | Term is a language unit on the level where precedence and associativity matters.
 nixExpr :: Parser NExprLoc
 nixExpr = keywords <|> nixLambda <|> nixExprAlgebra
  where
--- a/src/Nix/Parser.hs
+++ b/src/Nix/Parser.hs
@@ -259,7 +259,7 @@ nixUri =
                 \ x ->
                   isAlphanumeric x
                   || (`elem` ("%/?:@&=+$,-_.!~*'" :: String)) x
-          pure $ NStr $ DoubleQuoted $ one $ Plain $ start `Text.cons` protocol <> ":" <> address
+          pure . NStr . DoubleQuoted . one . Plain $ start `Text.cons` protocol <> ":" <> address
 
 
 -- ** Strings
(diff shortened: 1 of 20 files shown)


src/Nix/Parser.hs
```diff
@@ -862,6 +863,8 @@ nixExprAlgebra =
       nixOperators nixSelector
     )
 
+-- | Nix term parser.
+-- | Term is a language unit on the level where precedence and associativity matters.
```
layus: This comment is at the wrong place.

tests/PrettyParseTests.hs
```diff
@@ -81,10 +81,14 @@ genAttrPath =
...
-      (Gen.list (Range.linear 0 10) $ liftA2 (,) asciiText $ Gen.maybe genExpr)
-      Gen.bool
+  , liftA3 (\ c b a -> ParamSet (pure $ coerce c) (Variadic `whenTrue` b) (coerce a))
       (Gen.choice [stub, asciiText])
+      Gen.bool
+      (Gen.list (Range.linear 0 10) $
+        liftA2 (,)
+          asciiText
+          (Gen.maybe genExpr)
+      )
```
layus, line 95: If the point is to keep the arguments order the same, would you accept to also rename c b a to a b c ? This code trades one confusion for the other :-).

tests/PrettyParseTests.hs
```diff
@@ -108,21 +108,25 @@ genExpr =
...
-          , (10, Gen.resize (Size (n `div` 2)) genLet)
-          , (10, Gen.resize (Size (n `div` 2)) genWith)
-          , (1 , Gen.resize (Size (n `div` 2)) genAssert)
+          , (2 , Gen.resize (sizeBy 3) genBinary)
+          , (3 , Gen.resize (sizeBy 3) genSelect)
+          , (20, Gen.resize (sizeBy 2) genAbs)
+          , (2 , Gen.resize (sizeBy 2) genHasAttr)
+          , (10, Gen.resize (sizeBy 2) genLet)
+          , (10, Gen.resize (sizeBy 2) genWith)
+          , (1 , Gen.resize (sizeBy 2) genAssert)
```
layus: Could go even further with `downscaleBy = Gen.resize . Size . div n` and `(1 , downscaleBy 2 genAssert)`, or even map downscaleBy on curried nested tuples like `(1 , (2, genAssert))`. Am I going too far ? :thinking: :smile:

tests/EvalTests.hs
```diff
@@ -354,10 +354,10 @@ case_directory_pathexists =
...
 
 -- jww (2018-05-02): This constantly changes!
--- case_placeholder =
---   constantEqualText
---       "\"/1rz4g4znpzjwh1xymhjpm42vipw92pr73vdgl6xs1hycac8kf2n9\""
---       "builtins.placeholder \"out\""
+case_placeholder =
+  constantEqualText
+      "\"/1rz4g4znpzjwh1xymhjpm42vipw92pr73vdgl6xs1hycac8kf2n9\""
+      "builtins.placeholder \"out\""
```
layus, line 359: Any reason why it would be more stable now than then ?

tests/PrettyParseTests.hs
layus: Ok, fixed :-)

Anton-Latukha:
You are right.

tests/PrettyParseTests.hs
layus: Oh, already outdated

tests/PrettyParseTests.hs
Anton-Latukha: Well, HNix is also a bit of a style practice for me, as Haskell grammar & inner workings are complex and weave between opposing forces. Style is still floating around to find "the place".

Since from the mathematic standpoint, things are written: `g . f`. I sometimes try to flip the `a, b c` trying to find the proper style. Maybe I should just pretend `g . f` math meme does not exist and use `a-z` name ordering.

At some time I should go into more low-level details of how GHC analyzes & compiles a function to understand the style derived from that.

But... HNix is open source. I do refactoring. In the meantime, I develop the coding style, as anything developing - it sometimes fluctuates a little.

src/Nix/Parser.hs
Anton-Latukha: Yes, I know about this place.

It is weird.

Would think a bit more about how to lay out the places around this better.

There is a conflict of `expression` definitions between the programming & parser terminology: https://markkarpov.com/tutorial/megaparsec.html#parsing-expressions (which is just an algebra inside a language), so the word `expression` is essentially (broken) taken away from us to be used here, so I try to explain what happens on several levels around this place without the reference to the term `expression`.

tests/EvalTests.hs
Anton-Latukha: `idk` :laughing:

If that commented-out test lived through all the time (3 years), & works regardless of the amount of reactors I did, and I enabled it and it works... the property does not look unstable to me.

I vaguely remember something around these, maybe they were Nix changes, remembering too vaguely.

We have no way of knowing (and I did not dig deep into it), because it was not explained why it was unstable, maybe there were technical difficulties during development, maybe it was forgotten to enable it after the thing resolved, maybe changes in the upstream, we do no know - that is why I just kept the memo.

Hash magically is the same, so the test technically works, when it would break - we would be able to provide a better explanation of why, if it is disabled.

tests/PrettyParseTests.hs
Anton-Latukha: It is not:

Screenshot-2021-10-14-13:51:46

The "outdated" is a purely technical label. (I think) GH generally doesn't hide the stuff until it seems to get addressed.

You are right.

I try to practice the PR review scoping more, generally, the best way to do the breather PR review for both sides is to address only the scope under the PR diffs, not above or around it. Additionally, everything probably should go with the word "please" to a contributor.

Like here - I did refactor, but you say "the more refactor can be done", which is true - but not the scope of the code given, I didn't write the code/form that was asked to refactor.

The refactoring is most effective to enforce on the initial code PR author. But ... it is a peculiar situation to tell a person "rewrite your own code in a better way", especially when they volunteer. So first merging the code, then somebody else refactoring it - seems like a lot easier natural strategy.

I generally like to do this stuff incrementally in the next PRs/commits.

Anton-Latukha: ... ah, yes, I already did that refactor already beforehand.

src/Nix/Parser.hs
Anton-Latukha: The dichotomy goes through that the function was initially called `nixExpr`, the parser definitions looks to be `term`, but the `nixTerm` already exists also, and how to properly call `nixTerm` - I do not even know, it just processes value units of a language, what is now `nixTerm` - is something around `lexing` and below `algebra`, `nixTerm` seems to be a `lexer`. Seems like parser terminology is conflicted, or I do not understand how it syncs-up with regular programming terminology.

Anton-Latukha: Removed the doc.

tests/EvalTests.hs
layus: I guess that the hash being stable over the last couple of years is proof enough that we should at least try to re-enable it. Thanks.

tests/PrettyParseTests.hs
layus: It makes sense, thanks for explaining your thought process.

src/Nix/Parser.hs
layus: Yep, naming is a hard thing :-).

tests/PrettyParseTests.hs
layus: Hi @Anton-Latukha,

I think your above comment deserves a more elaborate answer from my part. 

> I try to practice the PR review scoping more, generally, the best way to do the breather PR review for both sides is to address only the scope under the PR diffs, not above or around it.

This guideline makes sense to me, but the case of refactoring is a bit different. As the only point of refactoring is to improve code, it is a matter of style rather than functionality. And appreciating style is a matter of taking the context into account. 

> Additionally, everything probably should go with the word "please" to a contributor.

I think this all boils down to cultural differences, as adding "please" in every comment would seem weird for me (but at the same time your expectation is consistent with the title of your issues). For me, it should be obvious that I have no way to coerce you to do something. I also have no specific expectation on you to take my comments into account.

Said comments are here to show that the code has been at least roughly read by someone else, and to keep interacting about this project. I find that there is nothing more depressing than putting a lot of efforts into something and get no feedback at all.

As much time as writing patches may take, reviewing them (especially heavy ones like this one) is also time consuming. I always hope that this form of contribution may lead to fruitful discussions and friendly interactions. 

So again, thank you for keeping the project alive :-)

Anton-Latukha: Yes, you are right.

About usage of suggestion - is just a remark, it is not towards persona or really "a thing" there, I just used a tiny thing as a tiny example to drive attention, to have a dialog about the frequent problem maintainers/reviewers do and which is perceived frequently. Sometimes maintainers/reviewers pull their power too hard, as the merge rights / review process is pretty totalitarian rules - it either {exists,fixed,done} or not - it is easily blockable/derailable process, as even the comment that points something out in the code without optionality degree - gets perceived as a requirement. As a contributor I also couldn't tell "yes - that is a bit too far", so as it was phrased from my culture - either I needed to do it, or ignore it - which would be rude, either respond "yes, lets do not get in the weeds", which would be also rude, I'd needed to use quite an imagination to get out of a double bind phrasing. So started to talk that we as maintainers/reviewers needs to be kind with words to explain & establish the humane & pleasant process, & to give flexibility to all of it. Any kind of ask should go with the kind asking words - it greases the wheels of any work.

I just wanted to talk a bit about that open source collaboration really works on denoting the degrees of the optionality of requests & on sharing kindness.

PR #988, opened 2021-10-16
Make `typeFun` total

mxxun:
Replaces `typeFun`'s `foldr1` with a total `foldr`.
Admittedly NonEmpty syntax is not as pretty.


Anton-Latukha:
Thank you.

PR #1011, opened 2021-11-14
fix getFreeVars and add test

Radvendii:
the treatment of variable bindings was of so e.g. `{x}: x` was said to have `x` as a free variable. This seemed to just be a mistake to me, so I fixed it and added a test for that and similar cases.


Anton-Latukha:
I remember some talk about `{x}: x` - but do not remember what the theme was. But your fix makes sense & I love things that make sense.

Anton-Latukha:
> Most probably, in C++ Nix - the check for `rev == ""` - suddenly initializes the `rev` with a value

Noop: https://github.com/haskell-nix/hnix/pull/1012

I still can not figure-out why Nix does it.

Anton-Latukha:
CI is rough currently.

I navigate on Nixpkgs pinned build.

---

Also tested it locally.

PR #1015, opened 2021-11-25
Fix nix-shell invocation

Anderssorby:
nix-shell passes the parameter `inNixShell` now

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -109,6 +109,7 @@
       else abort "Requires Nix >= 2.0"
 
 , mkDerivation   ? null
+, inNixShell ? false
 }:
 
 let


Anton-Latukha:
Lately, I got lazy tracking the breaking changes that Nix introduces without any notification.

For example, `inNixShell` is not mentioned in: https://discourse.nixos.org/t/nix-2-4-released/15822
Not in: https://nixos.org/manual/nix/stable/release-notes/rl-2.4.html
Not in: https://nixos.org/manual/nix/stable/release-notes/rl-2.3.html
& https://discourse.nixos.org/t/nix-2-3-5-manual-and-release-notes/7551

& so, because I currently have no time to be enough in Nixpkgs to hear about implicit undocumented changes - so.

I've heard that `nix shell` now `inNixShell`, but not had covered completely what that means.
  
Thank you :)

Anton-Latukha:
At once I would say (or you already know) - Nix cuts corners & skews typing rules - it is a challenging task to proof Nix lang code - you would basically need to disregard some of Nix behaviors - if you would find such - I would be thankful for growing the: https://github.com/haskell-nix/hnix/blob/d3d4356dd2588427a3a24fc8776aca42b472c472/tests/NixLanguageTests.hs#L66-L72

list.

I would be grateful if HNix would grow the list of C++ implementation quirks to not copy.

Nix submodule can be updated & there is a bunch of additional tests there, which need to be sorted.

Issue #1000, opened 2021-11-02
Nix 2.4 changed escaping


Anton-Latukha:
Change is not documented.

https://github.com/haskell-nix/hnix/runs/4081568140?check_suite_focus=true

```
   Eval comparison tests
    ind-string-14.nix:                                               FAIL (0.02s)
      tests/TestCommon.hs:68:
      tests/eval-compare/ind-string-14.nix
      expected: "\"Escaping of ' followed by ': ''\\nEscaping of $ followed by {: \\${\\nAnd finally to interpret \\\\n etc. as in a string: \\n, \\r, \\t.\\n\"\n"
       but got: "\"Escaping of ' followed by ': ''\\nEscaping of $ followed by {: ${\\nAnd finally to interpret \\\\n etc. as in a string: \\n, \\r, \\t.\\n\"\n"
      Use -p '/ind-string-14.nix/' to rerun this test only
      
      tests/TestCommon.hs:68:
      tests/eval-compare/ind-string-05.nix
      expected: "\"  The \\\\ is not special here.\\n' can be followed by any character except another ', e.g. 'x'.\\nLikewise for $, e.g. $$ or $varName.\\nBut ' followed by ' is special, as is $ followed by {.\\nIf you want them, use anti-quotations: '', \\${.\\n\"\n"
       but got: "\"  The \\\\ is not special here.\\n' can be followed by any character except another ', e.g. 'x'.\\nLikewise for $, e.g. $$ or $varName.\\nBut ' followed by ' is special, as is $ followed by {.\\nIf you want them, use anti-quotations: '', ${.\\n\"\n"
      Use -p '/ind-string-05.nix/' to rerun this test only.
```

Anton-Latukha:
We hit #1000.

Recently I was doing a lot of work.
& now suddenly upstream changes breakages piled on.

I would rest & work at my own pace.

Anybody who wants can take this.

Anton-Latukha:
Report is a part of the #996

Anton-Latukha:
Seems like this function: https://github.com/NixOS/nix/blob/02dff9e5299eccd8199dfbf1b5c20bf6ed9f07bd/src/nix/get-env.sh#L104

Anton-Latukha:
It is probably some change regarding `Nix.Parser : nixString' :: Parser (NString NExprLoc)`

HNix code is complex. But complex in a much better meaning.

[...]

PR #1019, opened 2022-01-04
Escape ${ in strings when printing Nix expressions

soulomoon:
Fix for #1000 
Following the fix in [NixOS/nix 4012](https://github.com/NixOS/nix/pull/4012)
* What is done: 
  * stop escaping a single `$` sign in double quoted string if not following by `{` when printing.
  * escape `${` to `\${` in double quoted string when printing.
  * update the tests for this behavior.
  * refactor to drop the usage of `show` in `printNix`, and unify some double string escape behavior for `prettyString` and `printNix`

* What is not working:
  * test eval-okay-ind-string is malformed since the nix submodule is outdated(the test violate  [NixOS/nix 4012](https://github.com/NixOS/nix/pull/4012)), however the test does work for [Nix 2.4 version of eval-okay-ind-string.exp](https://raw.githubusercontent.com/NixOS/nix/2.4/tests/lang/eval-okay-ind-string.exp), this test should be banned until modern Nix submodule is pulled #610

### DIFF
--- a/src/Nix/Pretty.hs
+++ b/src/Nix/Pretty.hs
@@ -104,10 +104,11 @@ prettyString (DoubleQuoted parts) = "\"" <> foldMap prettyPart parts <> "\""
  where
   -- It serializes Text -> String, because the helper code is done for String,
   -- please, can someone break that code.
-  prettyPart (Plain t)      = pretty . foldMap escape . toString $ t
+  prettyPart (Plain t)      = pretty . dollarEscape . toText . foldMap escape . toString $ t
   prettyPart EscapedNewline = "''\\n"
   prettyPart (Antiquoted r) = "${" <> withoutParens r <> "}"
   escape '"' = "\\\""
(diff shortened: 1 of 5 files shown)


soulomoon:
Then I have done a refactoring of `escape :: Char -> [Char]` because I saw the comment.
Since we need to identify two Char pattern, the work around making the code ugly.

There is a way to refactor the code into nicer version by doing explicit recursion using bidirectional pattern synonym.

The idea is coming from [here](https://stackoverflow.com/questions/4056449/haskell-bytestrings-how-to-pattern-match)

Anton-Latukha:
Checking the code locally for now (upstream shenanigans are currently not allowed to make GHC 9.0 CI usable).

Anton-Latukha:
Reviewed the code.

Works locally.

Anton-Latukha:
Thank you for this

Issue #966, opened 2021-07-08
Please, bring changes so the newFailingTests are no more


Anton-Latukha:
Logical easiest progression: `eval-okay-types -> eval-okay-hash -> eval-okay-hashfile`

2021-07-08 list:
https://github.com/haskell-nix/hnix/blob/cc7a472fddfa25015f77efd9ece9f98dc2705f37/tests/NixLanguageTests.hs#L57-L68

:zzz: : `eval-okay-fromTOML` is of last concern. It used (as expected) only a couple of times in Nixpkgs.

---

   - [x] `eval-okay-hash` (see https://github.com/haskell-nix/hnix/issues/966#issuecomment-1005656228)
   - [x] `eval-okay-hashfile` (#1021)
   - [ ] `eval-okay-path`
   - [x] `eval-okay-types` (#1020)
   - [x] `eval-okay-context-introspection` (#971)

soulomoon:
* `eval-okay-types` should have been fixed by #1020.
* `eval-okay-hash.exp` have been empty by the upstream [here](https://github.com/NixOS/nix/commit/3f192ac80ca421888c668896b63485486e1397ae#diff-925e3d8edc395af2213c250589bcc444d88f00c0c8f692b1b86d5a7237067b58) and `eval-okay-hash.nix`  is missing.
we should skip it. 
* `eval-okay-types` should have been fixed by #1021.

Anton-Latukha:
Ok, #128 addressed the `eval-okay-path`.

Thanks, you closed this report.

soulomoon:
I am glad that I helped.

PR #1021, opened 2022-01-05
tests: NixLanguageTests: enable test 'eval-okay-hashfile'

soulomoon:
add builtins.hashfile should suffice to pass eval-okay-hashfile
The main agenda & details are in the #966. This PR covers a part of it.


Anton-Latukha:
Looks normal.

If we would need more details to it, this is probably a nice groundwork.

Anton-Latukha:
Checked it locally.

Issue #1030, opened 2022-01-17
Meta: add builtins


Anton-Latukha:
Which are left:
Important:
  * [ ] #1055
  * [ ] #1034
  * [ ] #233
    Parts of implementation of it should be in #755
  * [ ] #240
  * [ ] #273
  * [ ] #162
  * [x] #128
    There is implementation of it in #755
    Ported in #1032
  * [x] Implement builtins.groubBy #1033
  * [x] Implement builtins.unsafeDiscardOutputDependency
    There is implementation of it in #755
    Ported in #1032
  * [x] Implement builtins.ceil #1031
  * [x] Implement builtins.floor #1031

Of smaller importance:
  * [ ] builtins.fetchMercurial
  * [ ] builtins.fromTOML
  
This is currently it.

soulomoon:
> Which are left: Important:
> 
> * [ ]  [Implement builtins.path #128](https://github.com/haskell-nix/hnix/issues/128)
>   There is implementation of it in [Support uploading filtered paths to the store #755](https://github.com/haskell-nix/hnix/pull/755)
> * [ ]  [Implement builtins.fetchTarball #162](https://github.com/haskell-nix/hnix/issues/162)
> * [ ]  [Implement builtins.filterSource #233](https://github.com/haskell-nix/hnix/issues/233)
> * [ ]  [Implement builtins.storePath #240](https://github.com/haskell-nix/hnix/issues/240)
> * [ ]  [Implement builtins.builtins #273](https://github.com/haskell-nix/hnix/issues/273)
> * [ ]  Implement builtins.groubBy
> * [ ]  Implement builtins.filterSource
> * [ ]  Implement builtins.storePath
> * [ ]  Implement builtins.unsafeDiscardOutputDependency
>   There is implementation of it in [Support uploading filtered paths to the store #755](https://github.com/haskell-nix/hnix/pull/755)
> * [ ]  Implement builtins.fetchGit
> * [ ]  Implement builtins.ceil
> * [ ]  Implement builtins.floor
> 
> Of smaller importance:
> 
> * [ ]  builtins.fetchMercurial
> * [ ]  builtins.fromTOML
> 
> This is currently it.

ceil and floor should be easy, and filterSource is duplicated in the list

Anton-Latukha:
Yes. That is why I kept them in "Important", because there is no point to postpone them.

Anton-Latukha:
[...]

Issue #880, opened 2021-03-11
Please, make the statement "Monoid is not []" true


Anton-Latukha:
One of the core stones HNix stumbles on is pretty pervasive String (as exceptions that gather provenance info, and in REPL) and, especially, List being used across the whole project.

The List is famously a one-directionally linked chain structure.

That means that per every cell used there is a cell that holds a pointer.
And there is no way to append to list without traverse of the whole list.

Appending `x` meaning traverse every time, which is doing `x^2` actions, so indeed - exponential append.

The most frequent thing the structure builder does is appending.

So, lets look at what actions happen with the Monoid, and most probably replace the List with an efficient Monoid, for example: https://hackage.haskell.org/package/acc

Anton-Latukha:
https://github.com/nikita-volkov/acc/issues/1#issuecomment-800976286

> *nikita-volkov:*
>
> There is a common pattern to construct data structures in two phases: first build, then freeze. With that in mind it helps to separate data structures into modification-optimized and frozen.
> 
> The first category is your builders, maps, sets, lists and what not, all optimized for various kinds of modification operations. Frozen types are optimized for querying operations and compact representation in memory, their examples are monolithic types like vectors, text, bytestring and ADTs around them.
> 
> Your `NExprF` and `NString` seem to fall into the frozen category. That's why I recommend using `Vector` instead of list or `Acc` in it.
> 
> Since `Vector` is not optimized for any sort of modification, you'll need a data structure optimized for building it. `Acc` paired with `Int` (for the vector size) would do for such purpose, however I've already released a package dedicated to building vectors and it is optimized even better for the task: ["vector-builder"](http://hackage.haskell.org/package/vector-builder).
> 
> I sense the way you construct is by parsing. It is a typical case for the build/freeze approach. More so, you'll find utilities like `many` and `some` in the "vector-builder" package that are particularly useful in parsing.

Anton-Latukha:
The codebase got refactored. It maximally uses polymorphic `one` for singletons & conversion to `NonEmpty` started. HLint guards for people not to put `[s]` singletons in code - so part of the work is done & protected.

It is now should be easier to convert `[]` to chosen types.

Anton-Latukha:
Use `OverloadedLists`:
https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/overloaded_lists.html

Anton-Latukha:
[...]

PR #1033, opened 2022-01-18
Builtins: add groupBy

soulomoon:
part of #1030

groupBy f list
* f could be NVBuiltin or NVClosure
* list should NVList

### DIFF
--- a/src/Nix/Builtins.hs
+++ b/src/Nix/Builtins.hs
@@ -33,10 +33,10 @@ import qualified "hashing" Crypto.Hash.SHA1    as SHA1
 import qualified "hashing" Crypto.Hash.SHA256  as SHA256
 import qualified "hashing" Crypto.Hash.SHA512  as SHA512
 import qualified Data.Aeson                    as A
-#if MIN_VERSION_aeson(2,0,0)
-import qualified Data.Aeson.Key                as AKM
-import qualified Data.Aeson.KeyMap             as AKM
-#endif
+
+
+
+
 import           Data.Align                     ( alignWith )
 import           Data.Array
 import           Data.Bits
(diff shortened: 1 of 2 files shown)


src/Nix/Builtins.hs
```diff
@@ -1435,6 +1435,37 @@ hashFileNix nsAlgo nvfilepath = Prim $ hash =<< fileContent
...
+   . MonadNix e t f m
+  => NValue t f m
+  -> NValue t f m
+  -> m (NValue t f m)
+groupByNix nvfun nvlist = do
+  list   <- demand nvlist
+  fun    <- demand nvfun
+  (f, l) <- extractP (fun, list)
+  mkNVSet mempty
+    .   fmap (mkNVList . reverse)
```
Anton-Latukha, line 1453: `reverse` sad face.

Currently *idk* how to reduce it here.

soulomoon: > `reverse` sad face.
> 
> Currently _idk_ how to reduce it here.

We can `flip (<>)` instead, either way is ugly (
[values in `fromListWith` is reversed](https://hackage.haskell.org/package/unordered-containers-0.2.14.0/docs/src/Data.HashMap.Internal.html#fromListWith)

Anton-Latukha: As far as I understand, `(<>)` for `[]` is worst-case `O(x^2)` (to the number of appends) (if every append is a snoc, it would be `O(x^2)`), while `reverse` `[]` of current `base` is (seems always) `O(x^2)` to the number of elements.

But this `reverse` seems to signal the pattern in data types. It needs to be looked into together with #880

soulomoon: Dig a little.
* The `reverse` should be fine, is O(n), done in recursion with accumulator, [here](https://hackage.haskell.org/package/base-4.14.3.0/docs/src/GHC-List.html#reverse).
* The `flip (<>)` would be worse, would result in `acc <> one x`, append to the end every encounter, O(n^2). 

as for #880, we do need a lot of reverse, if we prepend in situation where append is needed for performance reason. And switching away list seems big change for the situation.

PR #1039, opened 2022-01-21
String: improve function names

Anton-Latukha:


soulomoon:
looking good to me, such rename make functionality more clear in such big project.

Anton-Latukha:
After the statement (in `0.18`), that I moved all getterts to `get*` suffix, - I still find getters that do not follow that style.

In the codebase there is a difficulty between that there are several `String*` & several `Context*` which are used close or in conjunction in the code and function names/getters did not entail the difference and it is hard to transmit the difference.

This PR currently is just a code save.

Issue #1041, opened 2022-01-21
Instead of function application being `NBinary NApp f arg`, it should be `NApp f arg`


Anton-Latukha:

Anton-Latukha:
Working on this one.

Issue #982, opened 2021-08-10
NOperatorDef lacks precedence & so OperatorInfo was invented because of that


Anton-Latukha:
It is strange that data type for operator definition has everything, except avoids having predefined absolute precedence for the operator.
https://github.com/haskell-nix/hnix/blob/57a55fd4874c3c8d07836367dce678d113f2f823/src/Nix/Parser.hs#L598-L604

I am certain - the OperatorInfo was invented as a side-car for it:
https://github.com/haskell-nix/hnix/blob/57a55fd4874c3c8d07836367dce678d113f2f823/src/Nix/Parser.hs#L718-L724

If we look at what is special (tail) between repetitive functions:
https://github.com/haskell-nix/hnix/blob/57a55fd4874c3c8d07836367dce678d113f2f823/src/Nix/Parser.hs#L724-L775

Also, since precedence is removed from the operation definition field - functions implicitly establish precedence by ordering pattern matches/constructors.

And if we de-shotgun-surgery mentioned above functions:
```haskell

getOperatorInfo :: _ -> OperatorInfo
getOperatorInfo = mmap spec
 where
  mmap = (m Map.!)
   where
    m =
      Map.fromList $
        concat $
          zipWith
            (concatMap . spec)
            [1 ..]
            l2
     where
      l2 :: [[(NOperatorDef, Operator Parser NExprLoc)]]
      l2 = nixOperators $ fail "unused"
  spec i =
    \case
      (NUnaryDef op name, _) -> [(op, OperatorInfo i NAssocNone name)]
      (NBinaryDef assoc op name, _) -> [(op, OperatorInfo i assoc name)]
      (NSpecialDef assoc op name, _) -> [(op, OperatorInfo i assoc name)]

```

All this 3 functions are there & all they do. They take an implicitly aligned `nixOperators` list, and (implicitly) treat its index as operator precedence. So function just takes the `NOperatorDef`, implicitly figures-outs the precedence from the Haskell hardcoded list & combines that info, with a loss of the info of the type of the operation.
[...]

PR #1042, opened 2022-01-21
Expr.Types: NExprF: add NApp

Anton-Latukha:
Implements #1041.


Anton-Latukha:
After this, the #982 craves to be done.

Issue #1045, opened 2022-01-22
Refactor: change PatternSynonyms in value module from Unidirectional to bidirectional


soulomoon:
In `Value.hs`
It is really cool to have the correspondence between nix and haskell.

These are how the conversion between nix and haskell currently are.

* Value constructor, `mkNV*`.
* Value destructor unidirectional PatternSynonyms, `NV*`.  

But it seems superfluous since these two can actually be merged together by switching the PatternSynonyms from unidirectional to bidirectional.
We would have unified constructor and destructor for values.

* pros, cleaner, coherent, shorter, easier to go through the code.
* cons, We now need both constraints in PatternSynonyms from each of the two parts, `(Comonad f, Applicative f)` to make it work.

I have a working draft here #1046, hope you guys could review it and give advices.

Anton-Latukha:
It poped up in discussions in the past.

I've not yet internalized the design semantics of all types of PatternSynonyms application in the code, but probably it is just the time to do so.

Liking the result in #1046.

soulomoon:
work is done for bidirectional patterns

PR #1046, opened 2022-01-22
Unify `mkNV*` and `NV*` patterns by bidirectional synonyms in value modules

soulomoon:
changes draft related to #1045

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,6 +9,8 @@
     * [(link)](https://github.com/haskell-nix/hnix/pull/1038/files) project was using `megaparsec` `{,Source}Pos` and to use it shipped a lot of orphan instances. To improve the situation & performance (reports [#1026](https://github.com/haskell-nix/hnix/issues/1026), [#746](https://github.com/haskell-nix/hnix/issues/746)) project uses `N{,Source}Pos` types, related type signatures were changed accordingly.
   * `Nix.Value`
     * Unify builder `mkNV*` and `NV*` patterns by bidirectional synonyms, a lot of builders `mkNV*` are removed, and merged to `NV*`. e.g. instead of builder `mkNVList`, `NVList` should be used.
+    * Constraint `NVConstraint f = (Comonad f, Applicative f)` was introduced, in order to unify builder `mkNV*` and `NV*` patterns.
+  
 
 ## [(diff)](https://github.com/haskell-nix/hnix/compare/0.15.0...0.16.0#files_bucket) 0.16.0
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,7 +8,7 @@
     * [(link)](https://github.com/haskell-nix/hnix/pull/1042/files) The central HNix type `NExprF` changed, the `NApp` was moved out of `NBinary` & now a `NExprF` constructor of its own, the type signatures were changed accordingly.
     * [(link)](https://github.com/haskell-nix/hnix/pull/1038/files) project was using `megaparsec` `{,Source}Pos` and to use it shipped a lot of orphan instances. To improve the situation & performance (reports [#1026](https://github.com/haskell-nix/hnix/issues/1026), [#746](https://github.com/haskell-nix/hnix/issues/746)) project uses `N{,Source}Pos` types, related type signatures were changed accordingly.
   * `Nix.Value`
-    * Unify `mkNV*` and `NV*` patterns by bidirectional synonyms, a lot of `mkNV*` are removed, and merged to `NV*`. e.g. instead of `mkNVList`, `NVList` should be used, same goes for .
+    * Unify builder `mkNV*` and `NV*` patterns by bidirectional synonyms, a lot of builders `mkNV*` are removed, and merged to `NV*`. e.g. instead of builder `mkNVList`, `NVList` should be used.
 
 ## [(diff)](https://github.com/haskell-nix/hnix/compare/0.15.0...0.16.0#files_bucket) 0.16.0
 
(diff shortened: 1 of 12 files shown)


src/Nix/Value.hs
```diff
@@ -266,6 +267,7 @@ hoistNValueF lft =
 
 -- * @__NValue'__@: forming the (F(A))
 
+type NvConstraint f = (Comonad f, Applicative f)
```
Anton-Latukha: Indeed `hnix` breaks the camelCase acronym rule during `NV` naming. "It was done such before me :laughing:", `NV` acronym is currently a kind of consistent convention, as in `NVBuiltin` or `NVValue` & others. It is a bit strange for `NV*` patterns to have `Nv*` named constraint.

If you would want to rename all `NV`s to `Nv` - that can be planned into the future, when the project would do package splitting transition, because it would include all downstream processes - it would be a moment to include such change in API then.

Because a lot of the API changes we do - downstream rarely uses, but the `NV -> Nv` move is massive in API.

---

This constraint would be used in a lot of places. So people would arrive to it & wonder why it is there. It would be useful to provide a minimal explanation to why it was needed, it is a nice place to explain both the bidirectional pattern solution and what this constraint does.

Something of a kind: "to allow/enforce bidirectionality in patterns".

src/Nix/Value.hs
```diff
@@ -398,66 +400,7 @@ unliftNValue' = hoistNValue' lift
...
-
-
--- | Haskell functions to the Nix functions!
-mkNVBuiltin' :: (Applicative f, Functor m)
-  => VarName
-  -> (NValue t f m -> m r)
-  -> NValue' t f m r
-mkNVBuiltin' name f = NValue' $ pure $ NVBuiltinF name f
-
-
```
Anton-Latukha: This would require at least a note in the changelog.

Currently, through 2 year period of work I understood that we currently can allow low eyeball level of logging (the tunning down of details & preciseness is seen in the [`ChangeLog` itself](https://hackage.haskell.org/package/hnix-0.16.0/changelog)), while the project restructures.
But still, when we are deleting such a big number of functions - the mention of it & how to switch n the log would be good.

Anton-Latukha (changes requested):

src/Nix/Value.hs
soulomoon: Thank you @Anton-Latukha  for pointing these problems out.

`NV` as short hand for Nix Value to have uppercase would make sense too.
`NVConstraint` does seem to be a proper name in the current situation.

And documentation definitely needed in such places. 
And I need to sort out the documentation for code changes.

src/Nix/Value.hs
soulomoon: Agree, I will add them to ChangeLog of 0.17.0

src/Nix/Value.hs
soulomoon: renamed to `NVConstraint` and adjust the doc done

src/Nix/Value.hs
Anton-Latukha: Indicated that is would be `0.17.0` , since changelog already has API breakages in the project.

Skipping versions is not possible (not advised in PVP), skipping major version would mean downstream would need to make a hole in the package ranges and would lead to ... etc...

So yes, `0.17.0` is the current entry.

soulomoon: sure, section added

src/Nix/Builtins.hs
```diff
@@ -126,14 +126,15 @@ instance
...
   toBuiltin name f =
-    pure $ mkNVBuiltin (coerce name) $ toBuiltin name . f <=< fromValue . Deeper
+    pure $ NVBuiltin (coerce name) $ toBuiltin name . f <=< fromValue . Deeper
 
 -- *** @WValue@ closure wrapper to have @Ord@
 
 -- We wrap values solely to provide an Ord instance for genericClosure
-newtype WValue t f m = WValue (NValue t f m)
+data WValue t f m where
+  WValue :: NVConstraint f => NValue t f m -> WValue t f m
```
Anton-Latukha: Used GADTs here to hide the constraint, or there is more implication?

Asking, because I know GADTs were slow in old GHCs (need to find new information on it) & here was free `newtype` before. If GADT saves only a couple of lines, it get used only here. (oh, removing the `language GADTs` extension declarations in files).

soulomoon: GADTs here was just used to impose the constraint which can be impose else where.
This would carry the context as dictionary along and would be extracted when pattern match on them, which would introduce space overhead.
Take performance into consideration, newtype would be better, converting it back to newtype.

soulomoon: done, remove some other redundant unused GADTs extensional declarations as well.

Anton-Latukha: Ok, thanks.

Anton-Latukha (approved):
Great changes.

Anton-Latukha:
Thank you, I love these changes.

soulomoon:
Thank you too @Anton-Latukha for the help of reviewing the code.

ChangeLog.md
```diff
@@ -3,10 +3,17 @@
 
 ## [(diff)](https://github.com/haskell-nix/hnix/compare/0.16.0...0.17.0#files_bucket) 0.17.0
 
+* Introduction:
+  * `Nix.Value`
+    * [(link)](https://github.com/haskell-nix/hnix/pull/1046/files) Constraint `NVConstraint f = (Comonad f, Applicative f)` was introduced, in order to unify builder `mkNV*` and `NV*` patterns.
+
```
Anton-Latukha, line 9: Would note that Breaking & Additional - generally enough. If it is not major (breaking) change - everything else is additional niceties information. First - is first aid network, second is a daily news network.

For example, this change does not belong into "Introduction", it can be thought to belong to Additional, but as it changes type signatures in a number of places - it is a Breaking change. As it adds Comonad or Applicative (literal declaration of having requirement for such type class instance existence when there was none before - is a breakage) somewhere when it was not before.

The first most important client type of ChangeLogs in Haskell projects - is downstream developers. & generally under the semantic agenda of supporting the update, people look into changelogs. So the most important and clear thing ChangeLog should do - is to have as many breaking changes described precisely as possible. That it is an introduction of something into the code - can be said in explanation of breaking changes.

soulomoon: I was originally put it into breaking, but it comes to me that it is jut short hands for `Comonad f, Applicative f`.
An introduction of short hand for the new constraint requirements, I think it is both introduction and breaking, sort of dangling in the middle. 

Since you remind me of the change logs focus is on breaking changes and reading experience of developers , move back to breaking does seem to make more sense.  Thanks, surely will modify it.

soulomoon: done in #1048

PR #1053, opened 2022-02-09
(#1053) Fix compilation with ghc 8.6.5

hamishmack:
This change fixes the following errors when building with GHC 8.6.5
```
[51 of 52] Compiling Nix.Reduce       ( src/Nix/Reduce.hs, /Users/hamish/iohk/hnix/dist-newstyle/build/x86_64-osx/ghc-8.6.5/hnix-0.16.0/build/Nix/Reduce.o )

src/Nix/Reduce.hs:98:18: error:
    Ambiguous occurrence ‘putStrLn’
    It could refer to either ‘Nix.Prelude.putStrLn’,
                             imported from ‘Nix.Prelude’ at src/Nix/Reduce.hs:23:1-28
                             (and originally defined in ‘Relude.Lifted.Terminal’)
                          or ‘Prelude.putStrLn’,
                             imported from ‘Prelude’ at src/Nix/Reduce.hs:26:1-56
                             (and originally defined in ‘System.IO’)
   |
98 |         liftIO $ putStrLn $ "Importing file " <> coerce path''
   |                  ^^^^^^^^

src/Nix/Reduce.hs:102:48: error:
    Ambiguous occurrence ‘show’
    It could refer to either ‘Nix.Prelude.show’,
                             imported from ‘Nix.Prelude’ at src/Nix/Reduce.hs:23:1-28
                             (and originally defined in ‘Relude.String.Conversion’)
                          or ‘Text.Show.show’,
                             imported from ‘Prelude’ at src/Nix/Reduce.hs:26:1-56
                             (and originally defined in ‘GHC.Show’)
    |
102 |           (\ err -> fail $ "Parse failed: " <> show err)
    |                                                ^^^^

src/Nix/Reduce.hs:343:28: error:
    Ambiguous occurrence ‘show’
    It could refer to either ‘Nix.Prelude.show’,
                             imported from ‘Nix.Prelude’ at src/Nix/Reduce.hs:23:1-28
                             (and originally defined in ‘Relude.String.Conversion’)
                          or ‘Text.Show.show’,
                             imported from ‘Prelude’ at src/Nix/Reduce.hs:26:1-56
                             (and originally defined in ‘GHC.Show’)
```

One way to reproduce this issue is with the haskell.nix [hix tool](https://github.com/input-output-hk/haskell.nix/blob/master/docs/tutorials/getting-started-hix.md):

You can install hix with:
```
nix-env -iA hix -f https://github.com/input-output-hk/haskell.nix/tarball/master
```

Then run this to run cabal build in a nix-shell:
```
hix-shell --argstr compiler-nix-name ghc865 --argstr configureArgs '--disable-benchmarks' --run 'cabal build lib:hnix'
```

If you are on an M1 Mac you will need to pass `--system x86_64-darwin ` as well (ghc 8.6.5 does not work on `aarch64-darwin`).

### DIFF
--- a/src/Nix/Reduce.hs
+++ b/src/Nix/Reduce.hs
@@ -22,7 +22,7 @@ module Nix.Reduce
 
 import           Nix.Prelude
 import           Control.Monad.Catch            ( MonadCatch(catch) )
-#if !MIN_VERSION_base(4,13,0)
+#if !MIN_VERSION_base(4,12,0)
 import           Prelude                 hiding ( fail )
 import           Control.Monad.Fail
 #endif


Anton-Latukha:
Because of the 3 major versions policy, the GHC 8.6.5 was really hanging by the thread. I kept it in the code for the release and currently, but the check from the CI was removed to follow the policy.

From the PR it is self-evident that GHC 8.6.5 support seems to be still in use.

What would be the short description of its use and how long that is?

hamishmack:
Haskell.nix still supports ghc 8.6.5.  It would be possible to treat ghc 8.6.5 as a special case and use an older version of hnix with it, but it is nicer if all the compilers work the same.  In particular it is nice to be able to build the latest [`nix-tools`](https://github.com/input-output-hk/nix-tools) (programs Haskell.nix uses to generate nix) with ghc 8.6.5.

Anton-Latukha:
Remember we were talking about some `haskell.nix` supporting questions several years ago.
Well, as I'd been glad to have projects "post-MonadFail", currently supporting GHC 8.6 is not such a big deal. Thankfully project stack does not depend on GHC API changes (as HLS for example).
Defacto extending the support should be Ok.

Since `haskell.nix` keeps appearing, I think that addendum of project stack to try to include `haskell.nix` support should be enough.

[nix-tools](https://github.com/input-output-hk/nix-tools) idea does seem interesting. I hope eventually IOHK or somebody would get the Cabal -> Nix design right, which it is - I have ideas, but everyone does & it is a lot of work, IOHK projects seem to be in the ballpark of the right approach.

Hoping that IOHK found the Nix tooling engineer to continue work in this direction.

PR #1061, opened 2022-03-10
add flake.nix

DavHau:


soulomoon:
There are a lot of options we can pass to `nix-build`.
e.g. network is not accessible in nix build and we should pass argument to stop the tests.

Under flake, how do we do the following `nix-build --arg value`?

DavHau:
Part of the idea behind flakes is to express all possible outputs as an attribute set. If there are different variations for your package, this should be expressed as several attributes. This restriction achieves benefits like evaluation caching etc.

There are probably only a limited number of combinations of your parameters that are relevant. Each of these should be exported via `packages.{some-name}`.

> e.g. network is not accessible in nix build and we should pass argument to stop the tests.

If network is not accessible during nix build, then why have an argument for that in the first place? These tests should probably always be disabled for the nix build.

sorki:
IMHO this looks good. Networking and tests is a separate concern.

Seems like there is a https://github.com/DeterminateSystems/update-flake-lock thingie that we could use with another action that tests flake based build.

DavHau:
> Seems like there is a https://github.com/DeterminateSystems/update-flake-lock thingie that we could use with another action that tests flake based build.

I do not see the necessity for such tool, except there is a good reason why you would want to constantly update your dependencies like nixpkgs.
I guess for most projects updating dependencies should be an intentional controlled action. This can be done manually by a developer by running `nix flake update` and then committing to the repo.

soulomoon:
Cool, we should move the build system step by step to nix flake. let us merge this first.
The remaining process would be record in #1071

PR #1084, opened 2022-12-30
deps: unordered-containers >= 0.2.14

kquick:
Set lower bounds for unordered-containers to match actual need.


kquick:
This HLint failure seems completely independent of the change I've proposed here, and I've no idea how to fix it.  Created https://github.com/rwe/actions-hlint-run/issues/260 to get more assistance.

Anton-Latukha:
Solutions ustream were discussed in: https://github.com/ndmitchell/hlint/issues/1208.

Here we had the problem by design, `hlint` version is pinned in the CI config,

Anton-Latukha:
Thank you.

PR #1091, opened 2023-10-08
Fix quasiquote free variable interpolation

ners:
Before this PR, HNix provided an implementation for interpolating free variables as Nix expressions. This was not called correctly, because the types did not match: it looked for `NExprLoc` instead of `NExpr`.

This PR fixes interpolation such that the following code works as expected:

```haskell
let foo = "foo" :: String
[nix|
{
  foo = foo;
}
|]
```
produces:
```nix
{
  foo = "foo";
}
```

To do this, we fix the `ToExpr` typeclass to generate `NExpr`. We also provide some additional instances.

The diff also includes some lines that have been reformatted when running Brittany on the source file.

### DIFF
--- a/src/Nix/TH.hs
+++ b/src/Nix/TH.hs
@@ -1,72 +1,92 @@
-{-# language QuasiQuotes #-}
-{-# language TemplateHaskell #-}
-
-{-# options_ghc -Wno-missing-fields #-}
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
Thank you for fixing it.

PR #1092, opened 2023-10-14
release 0.17.0

ners:
This PR bumps the version of HNix to 0.17.0.

In addition, it
- updates version bounds of several dependencies to make them compile with library versions as provided in nixpkgs
- fixes a few instances that were broken with GHC 9.6 and above, in a way that works on all versions
- addresses all the compile warnings that were printed when built with all the GHC versions
- adds GHC 9.6 to Github workflows

I have tested building this with GHC 9.0.2, 9.2.8, 9.4.6, and 9.6.3.


ChangeLog.md
```diff
@@ -1,6 +1,8 @@
 
 # ChangeLog
 
+## [(diff)](https://github.com/haskell-nix/hnix/compare/0.17.0...0.18.0#files_bucket) 0.18.0
+
 ## [(diff)](https://github.com/haskell-nix/hnix/compare/0.16.0...0.17.0#files_bucket) 0.17.0
```
Anton-Latukha: The link to `0.17.0` diff is already present in the `main` development branch. Releasing `0.17.0` with `ChangeLog.md` which proclaims `0.18.0` would confuse people who would refer to it as a source of data.

This one, please clean this (delete `0.18..0` entry so on `Hackage` release `ChangeLog.md` would be right for `0.17.0`). Waiting on this to merge PR.

src/Nix/Value.hs
```diff
@@ -142,6 +141,14 @@ data NValueF p m r
...
+  (==) (NVConstantF x) (NVConstantF y) = x == y
+  (==) (NVStrF      x) (NVStrF      y) = x == y
+  (==) (NVPathF     x) (NVPathF     y) = x == y
+  (==) (NVListF     x) (NVListF     y) = x == y
+  (==) (NVSetF  _   x) (NVSetF _    y) = x == y
+  (==) _               _               = False
+
 instance Eq1 (NValueF p m) where
   liftEq _  (NVConstantF x) (NVConstantF y) = x == y
   liftEq _  (NVStrF      x) (NVStrF      y) = x == y
```
Anton-Latukha: Great, fully lawful `Eq`.

Notice here, that there was a `-- ** Eq1`, which is a sectioning of the code, that helps to partition, organize, have the high view, and navigate through the code as a tree structure:

This module code structure is:
image

I believe this is supported in several editors nowadays. I used Emacs [`outshine`](https://github.com/alphapapa/outshine) at the time.
 
This is not that important to fix it to merge the PR, but to have a note that those are the code section denotations.

src/Nix/Value.hs
```diff
@@ -717,8 +724,12 @@ instance MonadDataErrorContext t f m => Exception (ValueFrame t f m)
 
 -- * @instance Eq1 NValue'@
 
+instance (Eq a, Comonad f) => Eq (NValue' t f m a) where
+    (==) (NValue' (extract -> x)) (NValue' (extract -> y)) = x == y
+
```
Anton-Latukha: The same thing, the section name was not noted.

hnix.cabal
Anton-Latukha, line 1: Thank you for doing the work on dependencies.

Anton-Latukha (changes requested):

Anton-Latukha:
@sorki @layus @expipiplus1 @Ericson2314, please, I believe I do not have all that is needed to do the release. Who can make the release, after this?

src/Nix/Value.hs
ners: Interesting. I personally prefer to do this sort of thing with code outline plugins based on syntax rather than comments. This way the code is the single source of truth.

I've added the relevant section names.

hnix.cabal
ners: Thank you for maintaining HNix! :heart:

sorki:
Looks like @sjakobi has access. I only have perms for `hnix-store-*`

Anton-Latukha:
Sorry for the time-out.

Thank you.

PR #1097, opened 2023-11-18
nix: compiler ghc8107 -> ghc947

sorki:


sorki:
Won't nix-build on 23.05 due to `lens-family-th` and `tempalate-haskell`. Might work on unstable.

PR #1100, opened 2023-11-21
GHC947, nixpkgs-unstable, nuke submodule

sorki:
Seems fine now.

The forked repo at https://github.com/haskell-nix/nix/ should be deleted, it has `forked from https://github.com/mightybyte/nix` instead of `NixOS/nix` and is no longer needed.

- The tests pass happily as well via `cabal test` in shell - `1 of 1 test suites (1 of 1 test cases) passed.`
- Same for `cabal haddock all`
- Both `nix-build` and `nix build` work


Anton-Latukha:
Closes https://github.com/haskell-nix/hnix/issues/671

PR #1104, opened 2024-01-04
cryptonite -> crypton

sorki:
`cryptonite` is unmaintaned, switch to `crypton` fork.

Similar to https://github.com/haskell-nix/hnix-store/pull/256


sorki:
Requires store release

